PR #10, opened 2020-05-20
fix: `/etc` mount, `--cri` flag path, and bump Falco to 0.23.0

leogr:
Fixes #9

As per https://github.com/falcosecurity/falco/issues/1203 and [the official documentation](https://falco.org/docs/running/#docker-privileged) (see `-v /etc:/host/etc:ro`)

/cc @nibalizer 

/kind bug

PR #11, opened 2020-05-21
remove ebpf restriction on mountEtcVolume

nathwill:
the [falco-driver-loader](https://github.com/falcosecurity/falco/blob/master/scripts/falco-driver-loader) script that is run by the falco container entrypoint invokes get_target_id() to detect the OS from /host/etc/os-release in both the load_kernel_module() and load_bpf_probe() paths, so gating host-etc mountPath on ebpf.enabled is breaking target detection in the load_kernel_module path.


nathwill:
yes, looks to be the case, thanks for the heads up @leogr, i will await the fix in #10.

leogr:
Btw thank you for having fixed this in [the current helm chart](https://github.com/helm/charts/pull/22474) :+1:

PR #15, opened 2020-05-24
wip: update daemonset.yaml (--disable-cri-async)

kris-nova:


falco/templates/daemonset.yaml
```diff
@@ -57,6 +57,7 @@ spec:
             - --cri
             - /run/containerd/containerd.sock
             {{- end }}
+            - --disable-cri-async # @kris-nova default to sync falco mode to eliminate syscall drops
```
victortrac, line 60: Instead of hard-coding it here, would it be more appropriate to let the user set this flag via `Values.extraArgs` (on line 66)?  Seems like the change here should be a documentation change that informs the user about the flag rather than a hard-coded value.

leogr:
More context here :point_right: https://github.com/falcosecurity/falco/issues/925

nibalizer:
@kris-nova Can you look in on this when you get a second?

leogr:
Since we have `extraArgs` here :point_right: https://github.com/falcosecurity/charts/blob/1769c50cccb10cd29a786395603fe849dceccad7/falco/templates/daemonset.yaml#L65-L67 I believe we can close this.
/close

If you think we still need further improvement, feel free to re-open this again!

PR #31, opened 2020-06-13
Charts releasing automation

### LINKED ISSUE #22: Charts releasing automation
**Motivation**

We just need to automate everything, as usual :smiley_cat: 
As also discussed during the last community call.

**Feature**

Automate the following tasks (eg. using CircleCI and/or bots):

- [ ] Each chart should have a way to trigger its release process that 
  - validates the chart (ie. checking the version, `helm lint`, etc..)
  - then runs `helm package <chart-name>` to produce the artifact
- [ ] Artifacts should be available here as single source-of-truth (eg. in the `gh-pages` branch)
- [ ] The helm repo index must be automatically updated after a new chart release comes in
- [x] Finally, everything must be published (currently using the gh-pages mechanism)
- [ ] Add this helm repo to hub.helm.sh https://github.com/falcosecurity/charts/issues/40

**Alternatives**

Currently, we have a [simple script](https://github.com/falcosecurity/charts/blob/master/release.sh) that automate part of this process. Anyway, that script builds the artifact on the developer machine, thus the release process is not fully transparent. 
I don't believe that is a real alternative. We can just stay with it until we find a better and more elegant solution.

**Additional context**

Some constraints:

- both `master` and `gh-pages` branches are protected
- We can have multiple charts inside this repo
- ~~We cannot use the GitHub release/tag feature (since the reason above)~~ solution found :point_right: https://github.com/falcosecurity/charts/pull/31
- I would like to keep the chart sources of some project in their own repositories (eg [falco-exporter](https://github.com/falcosecurity/falco-exporter/tree/master/deploy/helm/falco-exporter))
  - The main reason here is to keep app sources and chart sources in the same place, that's useful to do things like [this](https://github.com/falcosecurity/falco-exporter/blob/11de13ac70383947f6701f20767043a63bbf61e3/Makefile#L13)

xantrix:
Signed-off-by: xantrix <x4ntrix@gmail.com>

/kind feature
Fixes #22 
Related #32 

This PR is largely inspired by the amazing work at https://github.com/codecentric/helm-charts.
After any changes at `Chart.yaml` files (only those present in chart dirs at level one), the CircleCI pipeline identifies changed charts since the latest tag and packages them (generate a .tgz file).
Then using the tool [Chart releaser](https://github.com/helm/chart-releaser):
* release chart artifacts to GitHub Releases named for the chart version.
* update the Helm repo `index.yaml` file for those releases and push it to GitHub Pages (gh-pages branch)

NOTE: `Chart.yaml` changes must include version change (e.g version: 1.1.8 -> 1.1.9)

TODOs:
- [x] add Falco context to get a personal access token (GITHUB_TOKEN) mandatory to push in `gh-pages` branch and use Chart releaser (CR_TOKEN)
- [ ] linting and testing of charts
- [ ] update README.md
- [ ] create release.md

/cc @leogr


leogr:
Thank you for taking care of that!!! ❤️

/kind build

.circleci/release.sh
```diff
@@ -0,0 +1,92 @@
...
+    mkdir -p .cr-release-packages
+
+    rm -rf .cr-index
+    mkdir -p .cr-index
+
+    echo "Identifying changed charts since tag '$latest_tag'..."
+
+    local changed_charts=()
+    local charts_path='*/Chart.yaml' # we consider only charts dirs at level one
+    readarray -t changed_charts <<< "$(git diff --find-renames --name-only "$latest_tag_rev" -- $charts_path | cut -d '/' -f 1 | uniq)"
```
leogr, line 47: I like the idea of releasing only when `Chart.yaml` has changed. :+1: 

What happens if the version has not been changed within `Chart.yaml`?  
I guess `cr` will return an error because of the already existing tag/release. If so, is that error clear enough to allow us to identify the issue quickly?

nibalizer: I think this is dope as well. I didn't know about `readarray`. It looks like we'd release on any metadata changes in the charts, but that's ok I suppose.

.circleci/release.sh
```diff
@@ -0,0 +1,92 @@
...
+    cr index -o "$GIT_USERNAME" -r "$GIT_REPOSITORY_NAME" -c "$CR_REPO_URL"
+
+    git config user.email "circleci@users.noreply.github.com"
+    git config user.name "circleci"
+
+    git checkout gh-pages
+    cp --force .cr-index/index.yaml index.yaml
+    git add index.yaml
+    git commit --message="Update index.yaml" --signoff
+    git push origin gh-pages
```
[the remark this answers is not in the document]
nibalizer: Can create and allow a 'robots' group? If not, I'd say the benefits of a fully automated pipeline outweigh the value of a protected gh-pages branch. No one should be making pull requests or pushing to that branch, and if anything weird happens we can have the robots completely rebuild it.

nibalizer:
This looks awesome. Thank you @xantrix!

.circleci/release.sh
leogr: Ideally, we'd release when the version has changed.

Release on any metadata changes would be ok too if `cr` was able to overwrite the GitHub release.

AFAIK `cr` is not able to overwrite an existing tag. Thus I expect an error when `Chart.yaml` was changed without bumping the version number (not 100% sure, maybe we can work it around somehow?)

xantrix: > I like the idea of releasing only when `Chart.yaml` has changed. 
> 
> What happens if the version has not been changed within `Chart.yaml`?
> I guess `cr` will return an error because of the already existing tag/release. If so, is that error clear enough to allow us to identify the issue quickly?

Yes, you guessed it. Without changing also the `version` key after a `Chart.yaml` change you'll get:
```bash
+ cr upload -o xantrix -r test-helm-repo
Error: error creating GitHub release: POST https://api.github.com/repos/xantrix/test-helm-repo/releases: 422 Validation Failed [{Resource:Release Field:tag_name Code:already_exists Message:}]
```
ATM Chart releaser does not support overwriting of an existing tag
https://github.com/helm/chart-releaser/issues/65
but it should be an edge case.

xantrix:
> This looks awesome. Thank you @xantrix!

I'm very happy you like it! :tada: 
Also, I'd like to thanks @leogr for being always helpful and responsive.

.circleci/config.yml
```diff
@@ -0,0 +1,29 @@
...
+            echo "export GIT_USERNAME=$CIRCLE_PROJECT_USERNAME" >> $BASH_ENV
+            echo "export GIT_REPOSITORY_NAME=$CIRCLE_PROJECT_REPONAME" >> $BASH_ENV
+            .circleci/install_tools.sh
+            .circleci/release.sh
+
+workflows:
+  version: 2
+  release:
+    jobs:
+      - release-charts:
```
leogr, line 24:
```suggestion
      - release-charts:
          context: falco
```
Add `context` to inject `GITHUB_TOKEN`

.circleci/release.sh
```diff
@@ -0,0 +1,92 @@
...
+
+release_charts() {
+    cr upload -o "$GIT_USERNAME" -r "$GIT_REPOSITORY_NAME"
+}
+
+update_index() {
+    cr index -o "$GIT_USERNAME" -r "$GIT_REPOSITORY_NAME" -c "$CR_REPO_URL"
+
+    git config user.email "circleci@users.noreply.github.com"
+    git config user.name "circleci"
```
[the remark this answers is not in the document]
leogr: As discussed with @leodido , @poiana can commit to a protected branch, so we can use our beloved bot.

.circleci/release.sh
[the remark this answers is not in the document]
leogr: @poiana should be able to commit to the protected branch, so this should be solved.

.circleci/release.sh
leogr: Solved by https://github.com/falcosecurity/charts/pull/31/commits/729562a57f4242c346101cb81b0fe7d5380d0c72

PR #33, opened 2020-06-15
Allow clientConfig url for AuditSink

jlsan92:
When pointing a `service` in the auditsink clientConfig, the scheme is always set to HTTPS.

I have a requirement to use only HTTP and these changes will allow that 👍. Seems that the only way to enforce it is by using `url` instead of `service` (even though my endpoint is in the same cluster)

I got inspiration from here: https://github.com/falcosecurity/evolution/blob/master/examples/k8s_audit_config/audit-sink.yaml.in


leogr:
Hi @jlsan92 

Thank you for your contribution!

Could you sign off your commit, please?

falco/values.yaml
```diff
@@ -89,6 +89,9 @@ auditLog:
   dynamicBackend:
     # true here configures an AuditSink who will receive the K8s audit logs
     enabled: false
+    # define if auditsink client config should point to a fixed url, not the
+    # default webserver service
+    url: ""
```
leogr, line 94: How can folks find the URL to put in here?

leogr, line 94: How can folks find the URL to put in here?

leogr: cc'ing @nibalizer @maxgio92 
WDYT?

maxgio92: I think that since the only part of the endpoint that would change would be the `scheme`, it could be specified whether to enable `plain` HTTP communication to the Falco Audit Backend, in which case:
- the `url` could be made up by the Falco service DNS name as so:
`http://<falco service name>.<falco namespace>:<falco listen port>/<falco audit path>`
- otherwise the `service` reference.

or did I miss something?

leogr:
> Probably we can refer to the official [docs](https://kubernetes.io/docs/tasks/debug-application-cluster/audit/#url)? By adding the link as the reference

It would help, but won't completely solve the issue, IMHO.
 
For example, in the [integration resources' README](https://github.com/falcosecurity/evolution/tree/master/deploy/kubernetes/k8s-audit-only#deploy-auditsink-objects) we suggested to do not use this approach in production because the DNS cannot be used, the service's IP may change, and the URL should not point to a service running in the cluster, as reported by the [official documentation](https://kubernetes.io/docs/tasks/debug-application-cluster/audit/#url):

> The host should not refer to a service running in the cluster; use a service reference by specifying the service field instead. The host might be resolved via external DNS in some apiservers (i.e., kube-apiserver cannot resolve in-cluster DNS as that would be a layering violation)

It's just a matter of explaining how to properly use this setting (eg. in the ["configuration" section of the chart's README](https://github.com/falcosecurity/charts/tree/master/falco#configuration)) in a way that users can avoid mistakes :)

That being said, I'm still convinced we need this.

Out of curiosity, if I may ask: Have you used a fixed IP with your use case? How do you get the IP?

Thank you again!

falco/values.yaml
leogr: AFAIK, the DNS cannot be used (see [here](https://kubernetes.io/docs/tasks/debug-application-cluster/audit/#url)) thus it cannot be automatically filled in by helm.

maxgio92: I was clearly missing something :)

leogr: Ok, I have added a note in the docs :point_right: https://github.com/falcosecurity/charts/pull/33/commits/02cdb50513212bb375c2e198ac3febd8bc58b914

PR #39, opened 2020-06-18
update(falco): update Falco rules from master

leogr:
/kind feature
/kind bug

This PR ports rule files from the current Falco master branch (https://github.com/falcosecurity/falco/commit/578ef7f64d94408fb2e07d30dd3f7131a054643c) to the charts.

N.B.: we will need to update those rules again when Falco 0.24.0 has been released.


leogr:
more updates are coming
/hold

antoinedeschenes:
Would be worth updating to the ruleset released with 0.23.0 at least, rules are missing changes from 0.22.0 at the moment. Should falco version bumps include the rules as well?

leogr:
Hi,

I'm unlocking this since it can be useful for folks that are currently experiencing issues with the default ruleset included in the helm chart.

/hold cancel

@antoinedeschenes 
I will update the rules again when Falco 0.24.0 has been released (in about two weeks), anyway you can yet pass the rule files you need by using the `falco.rulesFile` [configuration](https://github.com/falcosecurity/charts/tree/master/falco#configuration).

nibalizer:
What's the status of this?

leogr:
> What's the status of this?

I wanted to add this ASAP, just because it includes several fixes we discovered in rules and it can work with 0.23.0
We still need to update the rules again once Falco 0.24.0 has been released (I will take care of that).

So I think this can be merged rn.

/hold cancel

PR #51, opened 2020-07-17
Add SCC to support Openshift

airadier:
**What type of PR is this?**

/kind feature

/kind chart-release

/area falco-chart

The daemonset fails to create the pods in Openshift environment due to the request of elevated privileges, like HostPath mounts, privileged: true, etc.

This PR adds a SCC resource which is only created in Openshift (detected via API capabilities in the chart), and that can be disabled with scc.create: false in the values.yaml

**Checklist**

- [x] Chart Version bumped
- [x] CHANGELOG.md updated

### DIFF
--- a/falco/CHANGELOG.md
+++ b/falco/CHANGELOG.md
@@ -3,6 +3,12 @@
 This file documents all notable changes to Falco Helm Chart. The release
 numbering uses [semantic versioning](http://semver.org).
 
+## v1.2.1
+
+### Minor Changes
+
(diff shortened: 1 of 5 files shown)


nibalizer:
@airadier thanks for this! We just made a release this morning so I'm hesitant to do another one right away. Are you able to run from git for a little while while we stack up some commits for the next release?

airadier:
No problem, we already patched 1.2.0 with the SCC, so we can wait until the next release. Thanks!

PR #55, opened 2020-07-22
fix dir mounting collision when both gRPC and nats are enabled

### LINKED ISSUE #54: Duplicate mount point
**Describe the bug**

Since v1.2.x, Falco pods are Crashlooping due to duplicate mountpoint error related to `/var/run/falco` 

```bash
kubectl -n falco -l app=falco describe pods

Events:
  Type     Reason     Age                      From               Message
  ----     ------     ----                     ----               -------
  Normal   Scheduled  <unknown>                default-scheduler  Successfully assigned falco/falco-zwbxh to master02
  Normal   Pulling    71s                      kubelet, master02  Pulling image "busybox"
  Normal   Pulled     69s                      kubelet, master02  Successfully pulled image "busybox"
  Normal   Created    69s                      kubelet, master02  Created container init-pipe
  Normal   Started    69s                      kubelet, master02  Started container init-pipe
  Normal   Pulling    68s                      kubelet, master02  Pulling image "sysdig/falco-nats:latest"
  Normal   Pulled     67s                      kubelet, master02  Successfully pulled image "sysdig/falco-nats:latest"
  Normal   Created    67s                      kubelet, master02  Created container falco-nats
  Normal   Started    66s                      kubelet, master02  Started container falco-nats
  Warning  Failed     2s (x8 over 68s)         kubelet, master02  Error: Error response from daemon: Duplicate mount point: /var/run/falco
  Normal   Pulled     <invalid> (x9 over 68s)  kubelet, master02  Container image "docker.io/falcosecurity/falco:0.24.0" already present on machine
```

**How to reproduce it**

Run the following command to install Falco with grpc enabled:

```
helm repo add falcosecurity https://falcosecurity.github.io/charts
helm repo update
helm install --namespace falco falco --set falco.grpc.enabled=true falcosecurity/falco
```

**Expected behaviour**

v1.1.10 is working fine.
`shared-pipe` and `grpc-socket-dir` are both pointing to the `/var/run/falco` directory.

```text
spec:
  volumes:
    - name: shared-pipe
      emptyDir: {}
    - name: grpc-socket-dir
      hostPath:
        path: /var/run/falco
        type: ''
...
  containers:
      volumeMounts:
        - name: shared-pipe
          mountPath: /var/run/falco/
        - name: grpc-socket-dir
          mountPath: /var/run/falco
```

**Environment**

Kubernetes v1.18.4

- Falco version:
v0.24.0
- System info:

- Cloud provider or hardware configuration: Hetzner Cloud
- OS: Ubuntu 20.04 Focal Fossa

- Kernel: Linux edge01 5.4.0-28-generic #32-Ubuntu SMP Wed Apr 22 17:40:10 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux

- Installation method: Kubernetes

leogr:
**What type of PR is this?**

/kind bug

/kind chart-release

/area falco-chart

This PR fixes the duplicate `/var/run/falco` mounting problem (when both gRPC and nats are enabled) by prefixing mounting point inside the container with `/host` for the gRPC Unix socket path.
Note that at the host level the mounting point is still equal to one specified by the config (ie. without the `/host` prefix)

Fixes #54 

**Checklist**

- [x] Chart Version bumped
- [x] CHANGELOG.md updated

### DIFF
--- a/falco/CHANGELOG.md
+++ b/falco/CHANGELOG.md
@@ -3,6 +3,12 @@
 This file documents all notable changes to Falco Helm Chart. The release
 numbering uses [semantic versioning](http://semver.org).
 
+## v1.2.2
+
+### Minor Changes
+
+* Fix dir mounting collision when both gRPC and nats are enabled
+
 ## v1.2.1
 
 ### Minor Changes
--- a/falco/Chart.yaml
+++ b/falco/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: falco
-version: 1.2.1
+version: 1.2.2
 appVersion: 0.24.0
 description: Falco
 keywords:
--- a/falco/templates/_helpers.tpl
+++ b/falco/templates/_helpers.tpl
@@ -73,3 +73,21 @@ Extract the unixSocket's directory path
 {{- .Values.falco.grpc.unixSocketPath | trimPrefix "unix://" | dir -}}
 {{- end -}}
 {{- end -}}
+
+{{/*
+Extract the unixSocket's directory path and add /host prefix
+*/}}
+{{- define "falco.unixSocketDirWithPrefix" -}}
+{{- if .Values.falco.grpc.unixSocketPath -}}
+/host{{- .Values.falco.grpc.unixSocketPath | trimPrefix "unix://" | dir -}}
+{{- end -}}
+{{- end -}}
+
+{{/*
+Add /host prefix to the unixSocketPath
+*/}}
+{{- define "falco.unixSocketPathWithPrefix" -}}
+{{- if .Values.falco.grpc.unixSocketPath -}}
+unix:///host{{- .Values.falco.grpc.unixSocketPath | trimPrefix "unix://" -}}
+{{- end -}}
+{{- end -}}
--- a/falco/templates/configmap.yaml
+++ b/falco/templates/configmap.yaml
@@ -187,7 +187,7 @@ data:
       enabled: {{ .Values.falco.grpc.enabled }}
       threadiness: {{ .Values.falco.grpc.threadiness }}
       {{- if .Values.falco.grpc.unixSocketPath }}
-      bind_address: "{{ .Values.falco.grpc.unixSocketPath }}"
+      bind_address: "{{ include "falco.unixSocketPathWithPrefix" . }}"
       {{ else }}
       bind_address: "0.0.0.0:{{ .Values.falco.grpc.listenPort }}"
       private_key: {{ .Values.falco.grpc.privateKey }}
--- a/falco/templates/daemonset.yaml
+++ b/falco/templates/daemonset.yaml
@@ -131,7 +131,7 @@ spec:
             {{- end }}
             {{- if .Values.falco.grpc.enabled }}
             {{- if .Values.falco.grpc.unixSocketPath }}
-            - mountPath: {{ include "falco.unixSocketDir" . }}
+            - mountPath: {{ include "falco.unixSocketDirWithPrefix" . }}
               name: grpc-socket-dir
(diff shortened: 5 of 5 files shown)


leogr:
/cc @nibalizer 
PTAL

Could someone here help me in testing this PR with NATS enabled?

nibalizer:
I can try to test this @leogr :)

nibalizer:
This worked for me when I tested it. It brings up a couple questions:

1) Why is NATS support a special case now? The sidekick supports it (and a million other things) without a need for a emptydir/fifo or [custom daemon](https://github.com/falcosecurity/kubernetes-response-engine/blob/master/falco-nats/main.go) or a [docker image still in the sysding namespace](https://hub.docker.com/r/sysdig/falco-nats). I think we could clean up the helm chart a bit by saying if you want nats use the sidekick.

2) The use of /host is I think maybe not the right choice. If we know the exporter or another grpc-unixsocket consumer will be deployed in the same pod as falco I think we could use [emptydir](https://kubernetes.io/docs/concepts/storage/volumes/#emptydir).  I don't think using the host's filesystem is a bad idea but using the specific path of (the host's) `/var/run/falco` doesn't seem right to me and assumes any consumer pod would mount in that directory. As a user I'd rather not mount in /var/run from my host if I didn't have to.

leogr:
> 1. Why is NATS support a special case now? The sidekick supports it (and a million other things) without a need for a emptydir/fifo or [custom daemon](https://github.com/falcosecurity/kubernetes-response-engine/blob/master/falco-nats/main.go) or a [docker image still in the sysding namespace](https://hub.docker.com/r/sysdig/falco-nats). I think we could clean up the helm chart a bit by saying if you want nats use the sidekick.

I totally agree, but unfortunately, I don't have any experience with NATS so I can't make a decision.
cc @fntlnz @leodido 

> 2. The use of /host is I think maybe not the right choice. If we know the exporter or another grpc-unixsocket consumer will be deployed in the same pod as falco I think we could use [emptydir](https://kubernetes.io/docs/concepts/storage/volumes/#emptydir).  I don't think using the host's filesystem is a bad idea but using the specific path of (the host's) `/var/run/falco` doesn't seem right to me and assumes any consumer pod would mount in that directory. As a user I'd rather not mount in /var/run from my host if I didn't have to.

There're different issues here. 

Regarding `/host` I can agree. I have chosen that path just because Falco is already mounting other host paths there. I know there's no formal convention about that, but it seems to be a *de facto* standard for Falco.

Secondly, since Falco can be installed either with a DaemonSet or in the host directly, `/var/run/falco` is the natural place a consumer should look for the Falco UnixSocket socket. Indeed, falco-exporter - that does NOT run the in the same pod since it's a different deployment - mounts `/var/run/falco` from the host for this reason.
That being said, I'm not sure this is the best approach, but it was natural to me because of the host installation use case.

> My request here is to move the path to something unique and unimportant on the base host or to use an emptyDir.

I'd take a different strategy: if we could modify the NATS configuration to do not use `/var/run/falco`, then I will just leave the gRPC configuration as-is, and I will not merge this PR (mainly because of the host level use case).

But, can we modify the NATS?

nibalizer:
I believe the NATS code in question is actually sysdig/falco code [here](https://github.com/falcosecurity/kubernetes-response-engine/tree/master/falco-nats). So we could modify that, yes.

leogr:
> I believe the NATS code in question is actually sysdig/falco code [here](https://github.com/falcosecurity/kubernetes-response-engine/tree/master/falco-nats). So we could modify that, yes.

I have to look into that.
cc @nestorsalceda

leogr:
I have looked again into this and @nibalizer reasons are enough to close this PR

PR #57, opened 2020-07-22
fix(falco): use correct path to containerd socket

nvanheuverzwijn:
**What type of PR is this?**
Use correct path for containerd socket.

/kind bug

/area falco-chart

--cri option does not seem to use the correct path. Containerd socket is located in `/host/run/containerd/containerd.sock` instead of `/run/containerd/containerd.sock`.


nvanheuverzwijn:
After reading documentation, this is normal, see
https://github.com/falcosecurity/falco/commit/6c0e5297fa6f49eb0327a79b34be1bc0c90c8b0b

Closing this.

PR #71, opened 2020-08-14
falco: k8s-audit: Add SSL support

### LINKED ISSUE #38: [falco] Enabling SSL for the embedded webserver is not possible / auditLog.dynamicBackend cannot work
**Describe the bug**

Since the chart does not provide an option to [enable the SSL for the webserver](https://github.com/falcosecurity/charts/blob/2abd4af82df0cad84c5bc1ec54357916a416b5b6/falco/templates/configmap.yaml#L151) and when pointing to a [service reference](https://kubernetes.io/docs/tasks/debug-application-cluster/audit/#service-reference) in the AuditSink `clientConfig`  the scheme is always set to HTTPS, the [AudiSink configuration](https://github.com/falcosecurity/charts/blob/2abd4af82df0cad84c5bc1ec54357916a416b5b6/falco/templates/auditsink.yaml#L20)  won't work.

**How to reproduce it**

Install falco with `auditLog.dynamicBacked` enabled:
```shell
helm install falco falcosecurity/falco
    --set auditLog.enabled=true --set auditLog.dynamicBackend.enabled=true
```

Then errors like the following are continuously emitted to the K8s api-server log:
```
Error in audit plugin 'dynamic_webhook' affecting 1 audit events: Post https://falco.default.svc:8765/k8s-audit?timeout=30s: http: server gave HTTP response to HTTPS client
```

**Expected behaviour**

Ability to enable the SSL on the webserver and no errors.

**Screenshots**

**Environment**

- Falco version: `0.23.0`
- System info:

- Cloud provider or hardware configuration:
- OS:

- Kernel:

- Installation method: `Helm`

**Additional context**

antoinedeschenes:
**What type of PR is this?**
/kind bug
/kind feature
/kind chart-release

/area falco-chart

- Add `falco.webserver.sslEnabled` config, enabling SSL support using the `certs` values from gRPC
- Document `certs` values

Fixes #38

Testing SSL support might require a docker image with libssl1.0 installed if it's not statically linked yet
For SSL, the certificate has to be valid for the loopback IP used in the API server config.

**Checklist**

- [x] Chart Version bumped
- [x] CHANGELOG.md updated

### DIFF
--- a/falco/README.md
+++ b/falco/README.md
@@ -102,6 +102,8 @@ The following table lists the configurable parameters of the Falco chart and the
 | `falco.webserver.k8sAuditEndpoint`              | Endpoint where Falco embedded webserver accepts K8s audit events                                                   | `/k8s-audit`                                                                                                                              |
 | `falco.webserver.listenPort`                    | Port where Falco embedded webserver listen to connections                                                          | `8765`                                                                                                                                    |
 | `falco.webserver.nodePort`                      | Exposes the Falco embedded webserver through a NodePort                                                            | `false`                                                                                                                                   |
+| `falco.webserver.sslEnabled`                    | Enable SSL on Falco embedded webserver                                                                             | `false`                                                                                                                                   |
+| `falco.webserver.sslCertificate`                | Certificate bundle path for the Falco embedded webserver                                                           | `/etc/falco/certs/server.pem`                                                                                                             |
 | `falco.programOutput.enabled`                   | Enable program output for security notifications                                                                   | `false`                                                                                                                                   |
 | `falco.programOutput.keepAlive`                 | Start the program once or re-spawn when a notification arrives                                                     | `false`                                                                                                                                   |
 | `falco.programOutput.program`                   | Command to execute for program output                                                                              | `mail -s "Falco Notification" someone@example.com`                                                                                        |
@@ -116,6 +118,9 @@ The following table lists the configurable parameters of the Falco chart and the
 | `falco.grpc.rootCerts`                          | CA root file path for the Falco gRPC server                                                                        | `/etc/falco/certs/ca.crt`                                                                                                                 |
 | `falco.grpcOutput.enabled`                      | Enable the gRPC output and events will be kept in memory until you read them with a gRPC client.                   | `false`                                                                                                                                   |
 | `customRules`                                   | Third party rules enabled for Falco                                                                                | `{}`                                                                                                                                      |
+| `certs.server.key`                              | Key used by gRPC and webserver                                                                                     | ` `                                                                                                                                       |
+| `certs.server.crt`                              | Certificate used by gRPC and webserver                                                                             | ` `                                                                                                                                       |
+| `certs.ca.crt`                                  | CA certificate used by gRPC, webserver and AuditSink validation                                                    | ` `                                                                                                                                       |
 | `nodeSelector`                                  | The node selection constraint                                                                                      | `{}`                                                                                                                                      |
 | `affinity`                                      | The affinity constraint                                                                                            | `{}`                                                                                                                                      |
 | `tolerations`                                   | The tolerations for scheduling                                                                                     | `node-role.kubernetes.io/master:NoSchedule`                                                                                               |
@@ -200,6 +205,7 @@ And this means that our Falco installation has loaded the rules and is ready to 
 
 ## Enabling K8s audit event support
 
+### Using scripts
 This has been tested with Kops and Minikube. You will need the following components:
 
 * A Kubernetes cluster greater than v1.13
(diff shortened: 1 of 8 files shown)


antoinedeschenes:
> Hey @antoinedeschenes
> 
> Thank you for this awesome PR 
> 
> Could you add a bit more context about the `libssl1.0-dev` issue?
> I have not found any relevant issues about that.

@leogr This is related to error `Could not create embedded webserver: null context when constructing CivetServer.` due to `libcrypto.so` missing in issue https://github.com/falcosecurity/falco/issues/860 

I'm running a custom image, but the `ln` fix might be enough to get it to work with the stock falco image.

leogr:
> @leogr This is related to error `Could not create embedded webserver: null context when constructing CivetServer.` due to `libcrypto.so` missing in issue [falcosecurity/falco#860](https://github.com/falcosecurity/falco/issues/860)
> 
> I'm running a custom image, but the `ln` fix might be enough to get it to work with the stock falco image.

Ok, so I believe we should fix that problem in Falco first :)
/hold
blocked by https://github.com/falcosecurity/falco/issues/860

antoinedeschenes:
Puttting this back to WIP as Dynamic Auditing is removed from k8s 1.19. SSL support in the chart is still a useful feature.

antoinedeschenes:
@leogr changed the PR scope, I still add webserver SSL support, but replaced dynamicAuditing with a nodePort. It's currently working on our clusters

leogr:
> @leogr changed the PR scope, I still add webserver SSL support, but replaced dynamicAuditing with a nodePort. It's currently working on our clusters

First of all, I apologize for letting this PR here for such a long time. I'm just waiting to get a consensus regarding this issue. 
And thank you again. Your effort is highly appreciated :+1: 

I believe some users might have security concerns regarding the NodePort solution. Anyway, it may still be acceptable if not enabled by default.

cc @falcosecurity/charts-maintainers WDYT?

antoinedeschenes:
@leogr For the security aspect, client cert validation might be something worth looking at on the webserver side (enabling `ssl_verify_peer` on civet should work). The webhook config allows to use a client certificate on the kube API server when TLS is setup.

Otherwise one has to setup their node firewall, security group or GlobalNetworkPolicy to limit this to local access only at the moment.

falco/README.md
```diff
@@ -282,3 +330,26 @@ helm install falco \
   --set-file certs.ca.crt=/path/to/ca.crt \
   falcosecurity/falco
 ```
+
+## Generating SSL certificates for webserver (optional) or gRPC:
```
nibalizer, line 334: It's required for grpc, can we massage this line so it is obvious that it's needed for grpc?

nibalizer:
I think I'd like to see the dynamic audit sink configuration not removed. I'm not sure which kubernetes/falco versions a single chart is expected to support but removing it may impact a user. It'll stop working in later kubernetes versions anyways so we can just set this for removal in 6 months or a year or something.

I like re-using the certs from grpc.

I'm not confident about the choice of nodeport - is that flexible enough ... some people won't want it or will want to use ingress.

antoinedeschenes:
@nibalizer Dynamic Auditing never worked in the chart, as SSL could not be enabled, and was required to work. https://github.com/falcosecurity/charts/issues/38 this PR originally fixed that problem (see https://github.com/falcosecurity/charts/compare/master...antoinedeschenes:dynamic-audit-ssl?expand=1)

Should we then:
- fix dynamicAuditing for >=1.13 to <=1.18 (supported versions)
- make NodePort "opt-in"
- add ingress support

falco/README.md
leogr: It's required for both, but they are different setups.
gRPC can be configured with mutual TLS; on the other hand, you can basically enable HTTPS in the webserver.

I believe the better thing to do here is just to remove this block and add pointers to the official documentation:
- for gRPC :point_right: https://falco.org/docs/grpc/#generate-valid-ca
- for the webserver :point_right: https://github.com/civetweb/civetweb/blob/master/docs/OpenSSL.md#creating-a-self-signed-certificate

Those links are already provided by the "configuration" section within the doc (see https://falco.org/docs/configuration/).
@antoinedeschenes @nibalizer WDYT?

leogr: PS
gRPC via unix socket does not require mTLS.

leogr:
@nibalizer @antoinedeschenes 
I can confirm the dynamicAuditing never worked (unless the user patched the auditsink with a fixed URL).
So I think removing it is still an option. We can also suggest manually deploy the auditsink as explained [here](https://github.com/falcosecurity/evolution/tree/master/deploy/kubernetes/kernel-and-k8s-audit#deploy-auditsink-objects). The example should work with a helm deployed Falco chart too.

falco/README.md
antoinedeschenes: The gRPC instructions seem to fit both use-cases (except the client cert, but that would be a good thing to add to falco), I used subjectAltName with an IP address in my example, but "CN=localhost" makes more sense.

antoinedeschenes:
Changes:
- disabled the NodePort by default
- Removed SSL documentation and linked to gRPC documentation in both cases, as we're sharing the certificates and require a CA (not working for a self-signed cert)

nibalizer:
If we know it never worked I'm all for removing it :)

leogr:
Alright. I believe @antoinedeschenes did everything possible to make this PR perfect, so a big thanks to him! :clap: 
I will give it a try ASAP :+1:

nibalizer:
I wasn't able to get this working:

```
Mon Nov 30 23:14:53 2020: Starting internal webserver, listening on port 8765 (SSL)
Mon Nov 30 23:14:53 2020: Runtime error: Could not create embedded webserver: null context when constructing CivetServer. Possible problem binding to port.. Exiting.
```

That's with self signed certs and this:

```
helm install x --set-file certs.server.key=ssl/server.key --set-file certs.server.crt=ssl/server.crt --set-file certs.ca.crt=ssl/ca.crt   --set auditLog.enabled=true  --set falco.webserver.sslEnabled=true  .   
```

I think I saw something about self-signed certs not working... is that related? Thoughts?

leogr:
> I wasn't able to get this working:
> 
> ```
> Mon Nov 30 23:14:53 2020: Starting internal webserver, listening on port 8765 (SSL)
> Mon Nov 30 23:14:53 2020: Runtime error: Could not create embedded webserver: null context when constructing CivetServer. Possible problem binding to port.. Exiting.
> ```
> 
> That's with self signed certs and this:
> 
> ```
> helm install x --set-file certs.server.key=ssl/server.key --set-file certs.server.crt=ssl/server.crt --set-file certs.ca.crt=ssl/ca.crt   --set auditLog.enabled=true  --set falco.webserver.sslEnabled=true  .   
> ```
> 
> I think I saw something about self-signed certs not working... is that related? Thoughts?
> 
> 
> 
> 

The issue should be falcosecurity/falco#860

However, at this point I believe it would be better to split this PR in two: one for the SSL and another one for nodePort solution.

PS
I apologise for making the review process of this PR so long and a bit cumbersome, but it's better to wait until we are 100% convinced, IMHO. I hope you all understand my concerns.

antoinedeschenes:
@nibalizer Yes, civet webserver has dynamic dependencies for libssl1.0 which are missing in the docker image.
You can test with my forked image including PR https://github.com/falcosecurity/falco/pull/1344 which has the required dependencies: `ghcr.io/kronostechnologies/falco:0.26.2-stock.1`

nibalizer:
Wow! 

I second @leogr's apologies and thanks for sticking with us through this contribution

antoinedeschenes:
@nibalizer @leogr Just extracted the NodePort in #159 and AuditSink removal in #160

leogr:
> @nibalizer @leogr Just extracted the NodePort in #159 and AuditSink removal in #160

Hey @antoinedeschenes 

I have approved https://github.com/falcosecurity/charts/pull/159 right now.

Could you remove the nodePort implementation from here and then rebase, please?

Thank you in advance

antoinedeschenes:
@leogr done for this one and #160

nibalizer:
This looks good except for the chart version bump (I'm not sure we really should have that check enabled, but we do right now) and a changelog entry.

antoinedeschenes:
Quick note: I'm still assuming AuditSink is removed here, while it could be fixed by adding the caBundle field

leogr:
> Quick note: I'm still assuming AuditSink is removed here, while it could be fixed by adding the caBundle field

Thank you for the note, I didn't consider that. Since I believe it's better to wait before removing the AuditSink (see https://github.com/falcosecurity/charts/pull/160#pullrequestreview-550014878), it would be better if the AuditSink works too.

However, it should already work with `clientConfig.service` when using SSL :point_down: 
https://github.com/falcosecurity/charts/blob/master/falco/templates/auditsink.yaml#L21-L26
shouldn't it?

Let me know, thanks!

antoinedeschenes:
Ok just reread the documentation https://v1-18.docs.kubernetes.io/docs/tasks/debug-application-cluster/audit/ it uses CA validation, so would only work when using a trusted CA without the caBundle field set. (TLS remains mandatory) I can add this back to the PR

antoinedeschenes:
Ok, just added the AuditSink SSL config back, the documentation doesn't seem very clear about the `service` config requiring SSL or not, but they do require the url scheme to be `https`, and the default service port is 443.

falco/templates/auditsink.yaml
```diff
@@ -24,4 +24,7 @@ spec:
         port: {{ .Values.falco.webserver.listenPort }}
         path: {{ .Values.falco.webserver.k8sAuditEndpoint }}
       {{- end }}
+      {{- if .Values.falco.webserver.sslEnabled }}
+      caBundle: {{ .Values.certs.ca.crt | b64enc | quote }}
+      {{- end}}
```
leogr, line 29: Thank you!

nibalizer (approved):

nibalizer:
@leogr do you know why this has a hold?

leogr (approved):

leogr:
> @leogr do you know why this has a hold? 

Originally because of falcosecurity/falco#860

However, although that's not fixed yet, I believe this PR can be merged anyway, so

/hold cancel

PR #89, opened 2020-09-14
Add an initContainer in the falco daemonset

### LINKED ISSUE #36: [falco] Allow extra init container
**Motivation**

Reporting here the original request asked in https://github.com/helm/charts/issues/22268 

**Feature**

Allow extra init container(s) to be added, as Grafana does https://github.com/helm/charts/pull/12343

**Alternatives**

**Additional context**

This feature can be used to run `falcosecurity/driverloader` as init container, then `falcosecurity/falco-no-driver`. 
It would help to implement https://github.com/falcosecurity/charts/issues/17

Fsero:
I cannot use one of the provided drivers and falco cannot compile in my
OS (flatcar) so i've built a container that explicitly have the needed
sources and build and drop the driver in the host in the right path.

That way in the falco chart i can set the FALCO_BPF_PROBE to that path
and use the falco-no-driver image to load it.

Note: If you don't want to build your own image for driver building
you can use driverkit and populate the config for your kernel version
and OS.

**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

/kind chart-release

/area falco-chart

> /area falco-exporter-chart

> /area falcosidekick

This PR adds an initcontainer section in the daemonset under driverbuilder key over values. The main use case is to provide a custom image in order to build the falco driver if you cannot use or compile a driver using the official falco image. One of the images that can be used for this is driverkit with the right configuration.

Fixes #36

**Checklist**

- [x] Chart Version bumped
- [x] CHANGELOG.md updated


falco/values.yaml
```diff
@@ -15,6 +15,16 @@ containerd:
...
+# if you cannot download or compile the driver provided by falco.
+# use this initContainer to build it with your own image.
+#
+# driverbuilder:
+#  initcontainer:
+#    - name: init-driverloader
+#      image: mydockerimage
+#      command: ["sh", "-c", "./buildmykit"]
+#      ... extra params to initContainer.
 resources:
```
leogr, line 28: Although I like the idea to have an init container, I think this doc comment is too vague here (since we are not providing any init container for that).

Thus I'd propose to promote the `initcontainer` object to the top-level (removing the `driverbuilder` node, as suggested in the previous comments) in order to make it more general. At the same time, I'd add a more specific example that explains to the user how to leverage this feature to solve the use case you mentioned. 

WDYT?

falco/templates/daemonset.yaml
```diff
@@ -196,15 +196,22 @@ spec:
                   name: {{ template "falco.fullname" . }}
                   key: gcp-credentials-data
       {{- end }}
-      {{- if (or .Values.integrations.natsOutput.enabled .Values.integrations.snsOutput.enabled .Values.integrations.pubsubOutput.enabled) }}
+      {{- if (or .Values.integrations.natsOutput.enabled .Values.integrations.snsOutput.enabled .Values.integrations.pubsubOutput.enabled .Values.driverbuilder.initcontainer) }}
```
leogr, line 199: Why not add just a general way to add an init-container (instead of putting it under the driverbuilder section)?
```suggestion
      {{- if (or .Values.integrations.natsOutput.enabled .Values.integrations.snsOutput.enabled .Values.integrations.pubsubOutput.enabled .Values.initcontainer) }}
```

falco/templates/daemonset.yaml
```diff
@@ -196,15 +196,22 @@ spec:
...
       initContainers:
+          {{- if .Values.driverbuilder.initcontainer }}
+{{ toYaml .Values.driverbuilder.initcontainer | indent 12 }}
```
leogr, line 202:
```suggestion
{{ toYaml .Values.initcontainer | indent 12 }}
```

leogr (changes requested):
Thank you for your contribution!

Overall, it seems good to me though I'd like to improve it a bit (see my comments below).
Also, I'd bump the version to `v1.4.0` since we're introducing a new feature.

Moreover, could you sign off your commits, please?

falco/values.yaml
leogr: Also please see https://github.com/falcosecurity/charts/issues/36 (another way to this would as well as Grafana did in its chart :point_right: https://github.com/helm/charts/pull/12343 )

Fsero: Hi @leogr thanks for the review :)

sounds good to me as well. Let me work on that and i'll come back to you, i created the driverbuilder  to mark the intent clearer and because in some cases an initcontainer was added already.

leogr: Great!

Yep, I just like the idea of "extraInitContainers" because it opens many possibilities, including the one you need :)

Fsero: @leogr i did your requested changes and tried it in my cluster and seems to work, can you take  a look again please?

Thanks!

leogr: It looks great! Thank you!!! :clap: 

I just need some time to test it, I will do that ASAP

falco/values.yaml
```diff
@@ -342,3 +342,47 @@ tolerations:
...
+#   - name: init-driverloader
+#     image: falcosecurity/falco-driver-loader:0.25.0
+#     env:
+#       - name: FALCO_BPF_PROBE
+#       - name: BPF_USE_LOCAL_KERNEL_SOURCES
+#         value: "true"
+#       - NAME: ENABLE_DOWNLOAD
+#         value: ""
+#     command: ["/bin/bash"]
+#     args: ["-c", "apt update; apt install -f -y; apt install -y flex bison; apt install -y gcc libssl-dev; falco-driver-loader; cp /root/.falco/falco-bpf.o /host/lib/modules/falco-bpf.o"]
```
leogr, line 368: Ok, I have found some time to try it.

Although the overall PR seems good to me, unfortunately, I found some issue here that needs to be addressed:

- the `falcosecurity/falco-driver-loader` should not need `apt update; apt install -f -y; apt install -y flex bison; apt install -y gcc libssl-dev;` to properly compile the driver, and that let me think there's a problem in that image
- I don't believe we should encourage folks to put the eBPF probe in `/host/lib/modules/falco-bpf.o`, but I understand we need a place to put it so the main container can use it

Thus, I just need to put this on hold until we found a solution, especially for the latter point.

However, it's not your fault! Thanks to you we have noticed these other issues! :+1:

leogr (approved):
Hey,

I have changed my mind a bit and I have decided to just remove the example, so we can unblock this PR.
We will fix that separately, see #92

Thanks again for your contribution!

/approve

PR #113, opened 2020-10-02
Fix pipe in table

czunker:
Here you can find an example which explains it:
https://github.github.com/gfm/#example-200

Signed-off-by: Christian Zunker <christian.zunker@codecentric.cloud>

**What type of PR is this?**

/kind bug

/kind documentation

Fixes table in README.md


leogr (approved):
Thank you! :clap:

PR #139, opened 2020-11-16
Fix falcosidekick configuration table to use full path of configuration properties

KeisukeYamashita:
**What type of PR is this?**

> /kind documentation

/area falcosidekick

Because the current configuration description is misleading because there are items that used by `.Values.XXX` and some are used by `.Values.config.YYY` in the template. When I first saw the document, I thought I should configured as below:

```yaml
debug: true

slack:
   ....
```

But the correct one is:

```yaml
config:
    debug: true
   
     slack:
         ...
```

The falco sidekick didn't fail and ended up running with no output and not in debug mode.
I think we should add the full path not to confuse the users.

**Checklist**

- [x] Chart Version bumped
- [x] CHANGELOG.md updated


leogr (approved):
Thank you!

Issif:
Thank you a lot @KeisukeYamashita 👍

PR #155, opened 2020-12-03
wip: Add option to use existing secrets for falcosidekick chart

### LINKED ISSUE #145: Allow use of existing secrets in sidekick chart
**Motivation**

We manage our kubernetes deployments using a gitops workflow, in practice this means as much of our config as possible is checked into git. Currently this is difficult with the falco sidekick chart because the secrets for the various notifications backends have to be put in a `values.yaml` file. Although it is possible to encrypt this file etc, it forces a particular workflow rather than allowing us to use our normal workflow for managing kubernetes secrets. It would be useful if we could take care of managing the secrets in our own way and just point the sidekick chart to the appropriate secret.

**Feature**

Add an `existingSecret` value to the falco sidekick chart which if set would be referenced instead of the default secret which is created from the values passed to the `config` object. This is similar to how many other helm charts manage the same problem - for example see the `existingSecret` paramenters in https://github.com/bitnami/charts/tree/master/bitnami/postgresql.

**Alternatives**

Expose an helm parameter which allows injecting environment variables into the container which could be used to configure the secret values from an existing secret.

**Additional context**

I don't want to use helm secrets for our gitops workflow, I prefer to use something like bitnami sealded secrets or the sops-operator for managing encrypted secrets.

cartyc:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

/kind chart-release

> /area falco-chart

> /area falco-exporter-chart

/area falcosidekick

This PR is a work in progress to add the option to use existing secrets in order to make the `values,yaml` more friendly to version control and GitOps related release processes.

Currently the existing secret is enabled only for the GCP Pub/Sub and can be used by adding the secret name to the `existingsecret` field in the yaml.

```
  gcp:
    existingsecret: "sidekick" # The name of Secret to get values from
    credentials: "" # The base64-encoded JSON key file for the GCP service account
    pubsub:
      projectid: "" # The GCP Project ID containing the Pub/Sub Topic
      topic: "" # The name of the Pub/Sub topic
      minimumpriority: ""
```
 Alternatively you can still use the current process by leaving this blank.  The goal would be add the option to target a secret for every service in the same way as this.

An alternate way to do this would be to have the `existing` field set for all of the configs and just manage one secret instead of toggling them individually. 

I think the former would be a more flexible approach but I can see the value the latter. If there is a better way to approach this I'd be open to implementing that as well :)

Fixes #145 

**Checklist**

- [ ] Chart Version bumped
- [ ] CHANGELOG.md updated


falcosidekick/values.yaml
```diff
@@ -156,8 +156,12 @@ config:
     minimumpriority: ""
 
   gcp:
+<<<<<<< HEAD
```
Issif: you're rebase or merge failed ^^

cartyc: Ya I just fixed that. I had been using a Mac and switched over to my Windows box and didn't use the right hotkey to save the file :)

Issif:
Hi @cartyc,

I think we can close your PR now, @cpanato worked on a bigger one which embeds your changes, see #163 

Thanks for help for having proposed theses changes. 👍

cartyc:
That's awesome to see them in there! Thanks @cpanato!! 

/close

PR #159, opened 2020-12-04
falco: add falco.webserver.nodePort

antoinedeschenes:
Signed-off-by: Antoine Deschênes <antoine@antoinedeschenes.com>

**What type of PR is this?**
/kind feature

/area falco-chart

Allows exposing the audit log webhook through a NodePort for audit log configuration

Fixes #

**Checklist**

- [ ] Chart Version bumped
- [x] CHANGELOG.md updated


leogr (approved):
Although I don't think this should be the recommended way to use the AuditLog (some users may have concerns about that), it's ok to introduce this option since it's not enabled by default.

I think we should document this together with other alternatives. 

PS 
I have updated the changelog.
This feature will be released on the next chart release, not now.

PR #182, opened 2021-02-04
[Falco] Add the option to disable async CRI metadata fetching

Annegies:
/kind feature

/kind chart-release

/area falco-chart

Enables us to use the `--disable-cri-async`. 

None

Fixes #

**Checklist**

- [x] Chart Version bumped
- [x] CHANGELOG.md updated


falco/values.yaml
```diff
@@ -177,6 +177,13 @@ falco:
   # for up to 2 seconds without reaching the timeout.
 
   output_timeout: 2000
+  # Disable asynchronous CRI metadata fetching.
+  # This is useful to let the input event wait for the container metadata fetch
+  # to finish before moving forward. Async fetching, in some environments leads
+  # to empty fields for container metadata when the fetch is not fast enough to be
+  # completed asynchronously. This can have a performance penalty on your environment
+  # depending on the number of containers and the frequency at which they are created/started/stopped
+  disableCriAsync: false
```
cpanato, line 186: nit: should be `disableCRIAsync`

dza89: Two letter abbreviations is both capital letters.
From three and up it's only the first in capital.

Couldn't find a good source but XMLHTTPRequest or xmlHttpRequest

Annegies: fair, renamed!

Annegies:
resolved comments

Annegies:
is this okay?

leogr (commented):
Hi @Annegies 

Thank you for your effort!

Although it seems good to me, however, adding that flag was already possible by using `extraArgs`:
https://github.com/falcosecurity/charts/blob/master/falco/templates/daemonset.yaml#L72

I'm wondering if there's a real benefit to introduce an explicit configuration value for that (which could be redundant).

Anyhow, I have no strong opinion. It will be ok for me if this PR gets approved.
Just awaiting a bit more, to hear your and @falcosecurity/charts-maintainers opinions.

Annegies:
Hi @leogr!

Completely missed that value.
in that case imo this PR is completely redundant, so i'll close it.

Thank you for pointing this out!

PR #196, opened 2021-02-26
docs(falco): init container

### LINKED ISSUE #92: Add an example of how to use `extraInitContainers` configuration.
**Motivation**

See #36 and https://github.com/falcosecurity/charts/pull/89#pullrequestreview-488882738

**Feature**

Add an example of how to use `extraInitContainers` configuration.

**Alternatives**

**Additional context**

We may need an extra mount to pass the driver from the init container to the Falco container.
Thanks to @Fsero for his PR #89 and the [initial example](https://github.com/falcosecurity/charts/pull/89/files/6ecaec0918dc6dea00753fd0e9be432f6881a360#diff-c075bcc0c10d27070497f11bed4f0cccR347) (that has been removed in favor of this issue).

### LINKED ISSUE #147: [falco-chart] add extra volume mount (for init containers)
**Motivation**

We already introduced a feature to allow additional init containers :point_down: 
https://github.com/falcosecurity/charts/blob/3ca80c71251d92d64348ef6f9b9206ed8bcb829d/falco/values.yaml#L347

An init container would enable the use distinct image for loading the driver, then running Falco. For example, one could:

- use `falcosecurity/falco-driver-loader` as an init container that installs the driver and then exits
- then use `falcosecurity/falco-no-driver` (which smaller than the default image and does not contain the build toolchain) as the main container 

Although that should be already possible for the kernel module, with the ebpf probe is not possible since currently there is no way to share `/root/.falco/` (where the probe is stored).

For example, with `/root/.falco` shared volume between the init container and the main container as an extra volume mount (e.g. an emptyDir), the workflow should look like:
- `falcosecurity/falco-driver-loader` runs as initcontainer and writes to `/root/.falco/falco-bpf.o`
- `falcosecurity/falco-no-driver` runs as main container and reads from `/root/.falco/falco-bpf.o`

**Feature**

Add an `extraEmptyDirMounts` config node that allows sharing empty volumes. 

**Alternatives**

Instead of a configurable node, we could add an emptyDir volume mount for `/root/.falco/` by default.

**Additional context**

[This](https://github.com/helm/charts/pull/12343/files#diff-f6b6a9292748c6ab1fbebe14438a7d9d18a495f2541f93d7f751df1484e5b94cR72) is a very similar feature implemented by the Grafana helm chart.

For some other refs, see https://github.com/falcosecurity/charts/issues/92

cc @perezjasonr

leogr:
**What type of PR is this?**

/kind documentation

/kind chart-release

/area falco-chart

Fixes #92
Fixes #147

**Checklist**

- [x] Chart Version bumped
- [x] CHANGELOG.md updated

### DIFF
--- a/falco/CHANGELOG.md
+++ b/falco/CHANGELOG.md
@@ -3,6 +3,10 @@
 This file documents all notable changes to Falco Helm Chart. The release
 numbering uses [semantic versioning](http://semver.org).
 
+## v1.7.7
+
+* Add a documentation section about using init containers
+
 ## v1.7.6
 
 * Correct icon URL
--- a/falco/Chart.yaml
+++ b/falco/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: falco
-version: 1.7.6
+version: 1.7.7
 appVersion: 0.27.0
 description: Falco
 keywords:
--- a/falco/README.md
+++ b/falco/README.md
@@ -6,11 +6,11 @@
 
 This chart adds Falco to all nodes in your cluster using a DaemonSet.
 
-Also provides a Deployment for generating Falco alerts. This is useful for testing purposes.
+It also provides a Deployment for generating Falco alerts. This is useful for testing purposes.
 
 ## Adding `falcosecurity` repository
 
-Prior to install the chart, add the `falcosecurity` charts repository:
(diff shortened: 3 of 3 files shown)


cpanato (approved):
/lgtm

PR #305, opened 2022-01-12
Add configuration value to run unprivileged with capabilities

bauerd:
Signed-off-by: Dominic Bauer <dbauer@gitlab.com>

**What type of PR is this?**

/kind feature

/area falco-chart

This PR introduces a `leastPrivilege.enabled` value that enables running Falco [least privileged](https://falco.org/docs/getting-started/running/#docker-least-privileged).

If enabled, the `CAP_{BPF,SYS_RESOURCE,PERFMON}` capabilities are assigned, and the containers run unprivileged.

https://gitlab.com/gitlab-org/cluster-integration/cluster-applications/-/issues/51

The changelog does not have an entry for 1.16.3, yet Chart.yaml is at 1.16.3 currently. I bumped to 1.16.4 and added a changelog entry.

**Checklist**

- [x] Chart Version bumped
- [x] CHANGELOG.md updated


leogr (commented):
Hey @bauerd 

Thank you so much for this PR. We missed this feature for a long time :pray: 

I haven't tried it yet, but after a quick look at your implementation, it SGTM. I just have a couple of questions.

Is the `leastPrivileged` option intended to be used in conjunction with the [init container](https://github.com/falcosecurity/charts/tree/master/falco#using-an-init-container) approach?
Or do we assume the user installed the driver by other means?
Basically, how can the driver be installed without `privileged: true`?
(it would be great to document that in the README)

I see you did not include `SYS_PTRACE`. It is included in [our documentation](https://falco.org/docs/getting-started/running/#docker-least-privileged), but I can't recall why it is needed. Could you explain why it's not needed in this case?

Same question for `hostPid`. I guess it is not needed since Falco can still access the host `/proc`?

bauerd:
Hey @leogr, thanks for having a look, great questions.

>Is the leastPrivileged option intended to be used in conjunction with the init container approach?

It depends on the driver. It won't work to load the kernel module unprivileged. But attaching the eBPF probe works with `CAP_BPF` and `CAP_PERFMON`.

So yes, good point, I should definitely note that `leastPrivilege.enabled` conflicts with loading the kernel module driver.

For the record, installting the chart from the master branch, the `falco-driver-loader` init script which should load the kernel module fails me on GKE on both COS 93 (5.10.68+) and COS 89 (5.4.144+). It attempts to wget a prebuilt kernel module that 404s (see [logs](https://gist.github.com/bauerd/763d560f3abdadaa49390e2665eca719)). Unrelated to this PR though.

>I see you did not include SYS_PTRACE. It is included in our documentation, but I can't recall why it is needed. Could you explain why it's not needed in this case?

My understanding is that only the [pdig userspace driver](https://github.com/falcosecurity/pdig) uses ptrace(2). However pdig does not appear to be supported by this chart. Hence no need for `SYS_PTRACE` I think?

bauerd:
Hi, don't worry and take your time :+1: For the record, I also [tested this on GKE](https://gitlab.com/gitlab-org/cluster-integration/cluster-applications/-/issues/51#note_808272801).

falco/templates/daemonset.yaml
```diff
@@ -60,7 +60,14 @@ spec:
           resources:
 {{ toYaml .Values.resources | indent 12 }}
           securityContext:
-            privileged: true
+            privileged: {{ not .Values.leastPrivileged.enabled }}
+            {{- if .Values.leastPrivileged.enabled }}
+            capabilities:
+              add:
+              - BPF
```
FedeDP, line 67: Note that this requires runc 1.0.0-rc93 onwards: https://github.com/opencontainers/runc/pull/2553/files.  
In older versions, this won't work.

leogr (commented):
quick update:

- I believe this can work only with ebpf (see my suggestion below)
- There's a conflict; the PR needs a rebase

cc @bauerd
I'm still testing this PR. Once done, I can fix the comment and do the rebase by myself (so no action is required by you). Then, I'll give my final approval :)

paulopontesm:
Can this be merged? Is there something missing?

leogr:
Quick update:

We have done some investigation and testing. In the end, we discovered that:
- this approach can currently work only with eBPF and kernel >= 5.8
- it also requires runc >= 1.0.0-rc93 (otherwise some capabilities won't work)
- `CAP_PTRACE` is required (even if Falco does not use ptrace) when using `--pid=host` (it allows to read from processes environ). That was needed only when `HOST_ROOT=/`, but it's not the case for the helm chart.

So, I will rebase this PR soon and likely give my approval.

I will keep you posted.

PR #314, opened 2022-03-09
Fix quoting issue in falco chart daemonset template

tspearconquest:
Signed-off-by: Thomas Spear <tspear@conquestcyber.com>

**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /kind chart-release

/area falco-chart

> /area falco-exporter-chart

> /area falcosidekick-chart

Some kubernetes instances may pass along the quotes in a YAML list when the list item does not start with the double quotes, and this syntax causes issues with the new node hostname check as described below.

The falco-no-driver container image doesn't contain a shell, and so any users who have configured their helm installation to use that image may encounter startup issues on Falco 0.31.1 because there is no parsing of the double quotes before they are sent to the falco process, which causes the new hostname check in 0.31.1 to fail.

This PR changes the command line arguments to match the syntax originally proposed in falcosecurity/falco#1671 wherein the double quoted variable for the node name is on its own line. Kubernetes and jq both seem agree with this configuration by removing the quotes before passing the argument along to the entrypoint.

### DIFF
--- a/falco/CHANGELOG.md
+++ b/falco/CHANGELOG.md
@@ -3,6 +3,10 @@
 This file documents all notable changes to Falco Helm Chart. The release
 numbering uses [semantic versioning](http://semver.org).
 
+## v1.17.3
+
+* Fix quoting around `--k8s-node`
+
 ## v1.17.2
 
 * Add `leastPrivileged.enabled` configuration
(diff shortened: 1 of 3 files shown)


Issif:
Hello, thanks for your PR.

To make it mergeable, can you also update the changelog, it's requested by the CI. Thanks

leogr (approved):
LGTM

Thank you!

/approve

PR #320, opened 2022-04-09
Make falco.httpOutput.url always take precedence. Fixes #280

### LINKED ISSUE #280: falcosidekick httpOutput.url configmap issue
1. falco and falcosidekick generate service name based on fullname helper template:
https://github.com/falcosecurity/charts/blob/770a308d6203a04de640bce16eadcddc4d2c30fd/falcosidekick/templates/_helpers.tpl#L18-L22 
(in most cases .Release.Name-.Chart.Name-falcosidekick)
in some cases when using falco chart as requirement for a master chart and supplying falcosidekick.enabled=true , Chart.Name may be different for the falco and falcosidekick templates. causing httpOutput.url to be not match the generated service name
2. https://github.com/falcosecurity/charts/blob/770a308d6203a04de640bce16eadcddc4d2c30fd/falco/templates/configmap.yaml#L195 gives prevents a simple workaround to first issue by ignoring httpOutput.url variable when using falcosidekick.enabled=true 
httpOutput.url should override the generated value from falcosidekick to allow workarounds

mac-chaffee:
**What type of PR is this?**

/kind bug
/kind chart-release

/area falco-chart
/area falcosidekick-chart

When including the Falco helm chart inside a parent chart, that messes with the way the falcosidekick Service name gets generated, which in turn means falco can't send notifications. A user might expect that they could override this behavior by setting `falco.httpOutput.url` but currently that value is ignored if falcosidekick is enabled.

This PR fixes that issue by always allowing `falco.httpOutput.url` to override the auto-generated URL.

Fixes #280

Quick way to test:

```
$ cd falco/
$ helm template . --set falcosidekick.enabled=true | grep -A 3 "http_output"
    http_output:
      enabled: true
      url: |
        http://RELEASE-NAME-falco-falcosidekick:2801

$ helm template . --set falcosidekick.enabled=true --set falco.httpOutput.url=http://customurl | grep -A 3 "http_output"
    http_output:
      enabled: true
      url: |
        http://customurl
```

**Checklist**

- [x] Chart Version bumped
- [x] CHANGELOG.md updated

### DIFF
--- a/falco/CHANGELOG.md
+++ b/falco/CHANGELOG.md
@@ -3,6 +3,10 @@
 This file documents all notable changes to Falco Helm Chart. The release
 numbering uses [semantic versioning](http://semver.org).
 
+## v1.17.5
+
+* Changed `falco.httpOutput.url` so that it always overrides the default URL, even when falcosidekick is enabled.
+
 ## v1.17.4
 
 * Upgrade to Falco 0.31.1 (see the [Falco changelog](https://github.com/falcosecurity/falco/blob/0.31.1/CHANGELOG.md))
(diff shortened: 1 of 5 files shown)


mac-chaffee:
CC @Issif

leogr (approved):
LGTM

/approve

PR #351, opened 2022-06-14
feat(falco): Allow disabling of Falco's HTTP output with Falcosidekick

marc-sensenich:
**What type of PR is this?**

/kind feature
/kind chart-release

/area falcosidekick-chart

There are cases when you'd like to deploy both Falco and Falcosidekick using
a single Helm release and not use the default HTTP output configuration for
Falco. The introduction of a flag in the falcosidekick block of Falco's Helm
Chart values allows for disabling HTTP output and using another output such
as program to perform delivery of alerts to Falcosidekick for example.

N/A

N/A

**Checklist**

- [x] Chart Version bumped
- [x] CHANGELOG.md updated


Issif:
I get your point, but it seems to be a very particular use case. The deployment of Falcosidekick chart as dependency of Falco chart is for easy installation, but if you need a more complex installation with more control, it makes sense to tune settings yourself. wdyt?

marc-sensenich:
@Issif it was for a specific use case and it makes more sense to break out the deployments for a more complex use case, closing this PR. Thanks for the response!

PR #374, opened 2022-07-25
Added `mounts.enforceProcMount` flag

PhilipSchmid:
**What type of PR is this?**

/kind bug
/kind feature
/kind chart-release

/area falco-chart

When Falco is deployed without the syscall data source enabled (e.g. as Deployment for K8s audit events), it still mounts `/proc` by default, which often violates security policies (e.g. PSPs). 

When `/proc` would only be mounted when it's really required, the Falco Pod would not require a separate security policy handling whenever Falco is only configured to receive K8s audit events (or events from other plugins). Therefore, I propose to only mount it, whenever `.Values.driver.enabled` is set to `true` (so basically "by default"). To override this new default behaviour, the `mounts.enforceProcMount` flag was added to override it in case there is an edge case where the `/proc` mount is required but the syscall data source is not enabled at the same time.

This PR is the follow-up of a brief discussion with @leogr and @alacuku [from Slack](https://kubernetes.slack.com/archives/CMWH3EH32/p1658476444433579).

**Checklist**

- [X] Chart Version bumped
- [X] Variables are documented in the README.md (not required in my opinion)
- [X] CHANGELOG.md updated

### DIFF
--- a/falco/CHANGELOG.md
+++ b/falco/CHANGELOG.md
@@ -3,6 +3,10 @@
 This file documents all notable changes to Falco Helm Chart. The release
 numbering uses [semantic versioning](http://semver.org).
 
+## v2.0.5
+
+* Mount `/proc` only when syscall data source is enabled (default). This behaviour can be overridden via `mounts.enforceProcMount` for edge cases where the `/proc` `hostPath` mount is required without having the syscall data source enabled at the same time.
+
 ## v2.0.4
 
 * Fix templating for init containers in pod-template.tpl
--- a/falco/Chart.yaml
+++ b/falco/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: falco
-version: 2.0.4
+version: 2.0.5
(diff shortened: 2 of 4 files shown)


leogr (commented):
Hey @PhilipSchmid 

Thank you for this PR.

I think we have to add the same _if_ condition here :point_down: 
https://github.com/falcosecurity/charts/blob/5e22d5c4476bc4b1f1f961bd813a4ab550d79d17/falco/templates/pod-template.tpl#L127-L128

PhilipSchmid:
> Hey @PhilipSchmid
> 
> Thank you for this PR.
> 
> I think we have to add the same _if_ condition here 👇
> 
> https://github.com/falcosecurity/charts/blob/5e22d5c4476bc4b1f1f961bd813a4ab550d79d17/falco/templates/pod-template.tpl#L127-L128

Yes, absolutely. Sorry about that. I've now added it.

leogr (approved):
LGTM thank you!

PR #382, opened 2022-07-27
update(falco): deploy falco in privileged mode when using ebpf probe

alacuku:
```
Signed-off-by: Aldo Lacuku <aldo@lacuku.eu>
```

**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

/kind chart-release

/area falco-chart

> /area falco-exporter-chart

> /area falcosidekick-chart

> /area event-generator-chart

The default configuration for the `ebpf probe` uses the following configuration values `.Values.driver.ebpf.hostNetwork=true` and `.Values.driver.ebpf.leastPrivileged=true`. 

`hostNetwork=true` is needed to enable the `eBPF JIT` at runtime for performance reasons. It deploys Falco in host network in order for Falco pods to have access to `/proc/sys/net/core/bpf_jit_enable` file. As far as i know, there is not such logic implemented in Falco or [`driver-loader-script`](https://github.com/falcosecurity/falco/blob/master/scripts/falco-driver-loader) that would enable the `eBPF JIT` at runtime. Hence, this PR proposes to set the default value for `hostNetwork` to false, leaving to the users the possibility to enable it.

`leastPrivileged=true`, deploys Falco following the `principle of least privilege`. It does so by using linux `capabilities` to grant only the necessary permissions needed by Falco. The `least privileged mode` needs a `kernel >= 5.8` and at the same time the `container runtime implementation` installed in the node needs to support the following linux `capabilities`:  **CAP_BPF, CAP_SYS_RESOURCE, CAP_PERFMON and CAP_SYS_PTRACE**.  Being falco deployed in least privileged mode by default when using the `ebpf probe` most of the time falco does not work properly. This PR changes the default values of `leastPrivileged=false` increasing the chances that Falco will work out of the box when deployed with the `ebpf probe`.

Fixes #

Let me know what do you think about this changes.

**Checklist**

- [x] Chart Version bumped
- [x] CHANGELOG.md updated

### DIFF
--- a/falco/CHANGELOG.md
+++ b/falco/CHANGELOG.md
@@ -3,6 +3,10 @@
 This file documents all notable changes to Falco Helm Chart. The release
 numbering uses [semantic versioning](http://semver.org).
 
+## v2.0.8
+
+* When using ebpf probe Falco is deployed in `privileged` mode instead of `least privileged`.
+
 ## v2.0.7
 
 * Fix templating for priorityClassName in pod-template.tpl
--- a/falco/Chart.yaml
+++ b/falco/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: falco
-version: 2.0.7
+version: 2.0.8
 appVersion: 0.32.1
 description: Falco
 keywords:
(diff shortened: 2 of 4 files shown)


leogr (approved):
Make sense :+1: 

Thank you, Aldo!

PR #384, opened 2022-07-28
feat(sidekick-helm-doc): Values in README generated by helm docs

Lowaiz:
Signed-off-by: Lyonel Martinez <lyonel.martinez@numberly.com>

**What type of PR is this?**

/kind documentation

/area falcosidekick-chart

Helm docs integration based on falco chart one's

None

It seems that `config.kafka.partition` and `config.kafka.messageformat` was useless, the sidekick doesn't document these params. and they don't appear in the config parser, so I deleted them


Issif:
Amazing, seems good to me, are you ready for an approval?

falcosidekick/README.md
```diff
@@ -140,307 +140,299 @@ The command removes all the Kubernetes components associated with the chart and
...
-| `webui.redis.storageClass`                  | Storage class of the PVC for the redis pod                                                                                                                                             | `""`                                                                                                       |
-| `webui.redis.storageSize`                   | Size of the PVC for the redis pod                                                                                                                                                      | `1Gi`                                                                                                      |
-| `webui.redis.tolerations`                   | Tolerations for pod assignment                                                                                                                                                         | `[]`                                                                                                       |
-
-
-Specify each parameter using the `--set key=value[,key=value]` argument to `helm install`. For example,
```
Issif, line 440: We lose this part, can you add it back please

falcosidekick/README.md
```diff
@@ -140,302 +140,302 @@ The command removes all the Kubernetes components associated with the chart and
...
+| webui.resources | object | `{}` | The resources for the web UI pods |
+| webui.securityContext | object | `{}` | Web UI container securityContext |
+| webui.service.annotations | object | `{}` | The web UI service annotations (use this to set a internal LB, for example.) |
+| webui.service.nodePort | int | `30282` | The web UI service nodePort |
+| webui.service.port | int | `2802` | The web UI service port dor the falcosidekick-ui |
+| webui.service.targetPort | int | `2802` | The web UI service targetPort |
+| webui.service.type | string | `"ClusterIP"` | The web UI service type |
+| webui.tolerations | list | `[]` | Tolerations for pod assignment |
 
 Specify each parameter using the `--set key=value[,key=value]` argument to `helm install`. For example,
```
Issif, line 440:
```suggestion
Specify each parameter using the `--set key=value[,key=value]` argument to `helm install`.
```

falcosidekick/README.gotmpl
```diff
@@ -0,0 +1,152 @@
...
+
+The command removes all the Kubernetes components associated with the chart and deletes the release.
+
+## Configuration
+
+The following table lists the main configurable parameters of the Falcosidekick chart and their default values. See `values.yaml` for full list.
+
+{{ template "chart.valuesSection" . }}
+
+Specify each parameter using the `--set key=value[,key=value]` argument to `helm install`. For example,
```
Issif, line 145:
```suggestion
Specify each parameter using the `--set key=value[,key=value]` argument to `helm install`.
```

alacuku:
Hey @Lowaiz, nice job! 

Just a comment about the go template file. Having the README.md file generated using the go template file adds an extra step when we need just to update the docs. We first need to modify the content in README.gotmpl file, generate the README.md file and only then we can render and check the format of our docs. I would avoid this extra step by just generating the values table in separated file that is accessible from the README.md file. What do you think?

Lowaiz:
As I use a lot of helm chart, I always prefer a README including values rather than an redirection. Furthermore, there is already one redirection from the root to each chart, so I really prefer to not add one for the user.

Having values in the README, force, a bit, the user to read these values, make them a bit more accessible and probably avoid the users to miss some information.
I guess this is a personal preference, but having a extra step to render the README is a better option, IMHO, than isolate the values in an extra file.

Lowaiz:
> Amazing, seems good to me, are you ready for an approval? 

Ready when you are !

Issif:
> As I use a lot of helm chart, I always prefer a README including values rather than an redirection. Furthermore, there is already one redirection from the root to each chart, so I really prefer to not add one for the user.
> 
> Having values in the README, force, a bit, the user to read these values, make them a bit more accessible and probably avoid the users to miss some information. I guess this is a personal preference, but having a extra step to render the README is a better option, IMHO, than isolate the values in an extra file.

I'm adding the READMEs are imported by https://artifacthub.io/packages/helm/falcosecurity/falcosidekick, so it's a good practice to have all values there

Issif (approved):
amazing :+1: 
thank you

PR #417, opened 2022-10-26
Fix default falco values and doc

santrancisco:
**What type of PR is this?**

/kind bug

/kind documentation

/area falco-chart

When upgrading our old helm chart (v1.19.4) to the new version of the helm chart (v2.2.0) in our environment with the new helm chart values, we ran into an issue where, upon digging into what caused the error, we found the default `init_config` for the k8saudit plugin is set to `null` instead of empty string thus falco could not start. 

We ran into the following issue when updating our helm chart:

```
Tue Oct 25 11:15:27 2022: Runtime error: Error reading config file(/etc/falco/falco.yaml): could not load plugins config: yaml-cpp: error at line 35, column 16: bad conversion. Exiting.
```

This only happens if users do not specify any "plugins" and use the default value provided by the chart.

### DIFF
--- a/falco/CHANGELOG.md
+++ b/falco/CHANGELOG.md
@@ -3,6 +3,10 @@
 This file documents all notable changes to Falco Helm Chart. The release
 numbering uses [semantic versioning](http://semver.org).
 
+## v2.2.1
+
+* Fixed default helm chart value `init_config` for plugin `k8saudit` in values.yaml.
+
 ## v2.2.0
 
 * Change the grpc socket path from `unix:///var/run/falco/falco.soc` to `unix:///run/falco/falco.sock`. Please note that this change is potentially a breaking change if upgrading falco from a previous version and you have external consumers of the grpc socket.
--- a/falco/Chart.yaml
+++ b/falco/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: falco
-version: 2.2.0
+version: 2.2.1
 appVersion: 0.33.0
 description: Falco
 keywords:
--- a/falco/README.md
+++ b/falco/README.md
@@ -204,7 +204,7 @@ services:
         protocol: TCP
 
 falco:
-  rulesFile:
+  rules_file:
     - /etc/falco/k8s_audit_rules.yaml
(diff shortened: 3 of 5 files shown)


alacuku:
Hi @santrancisco, the Falco chart 2.0.0 introduces a breaking change since the labels used for the k8s objects have changed. Are you sure that helm allows you to do a `helm upgrade`? It should fail, causing the Falco deployment to be in an inconsistent state.

santrancisco:
Hi @alacuku, Apology for not being clear earlier, It was not a `helm upgrade` run but rather a new deploy with new helm chart values as we noticed the breaking changes. I was wondering how it wasn't a problem in the past and found [this commit](https://github.com/falcosecurity/charts/commit/8b77e1054cc498c7e2a21aa125aca23f24699435) that it may have been accidentally removed 4 months ago? 

CleanShot 2022-10-27 at 02 32 10

alacuku:
That's strange. I don't have issues when installing Falco with the new chart. Can you please provide the command you are running to install Falco?
Anyway, we still need the fix, could you please sign-off the commits, update the CHANGELOG.md and bump the chart version? So we can merge it!

santrancisco:
Updated :) Thanks @alacuku  for taking a look.

alacuku:
Hey @santrancisco, last thing, could you please run a `make docs` from the falco folder?

santrancisco:
@alacuku , mystery solved... after running make doc and saw the default variable in the doc changed from null to "",  I thought i would tweak the falco.yaml on a nix machine with falco installed to reproduce the behaviour anyway... Is it turned out, the latest falco version 0.33 made some changes (i have not digged into the code) and with the same falco.yaml config file I saw 0.31 and 0.32 version both threw a "bad converstion" error while 0.33 does not. Since you just recently updated the Chart with the latest version of falco (0.33),  I dont think we need this tiny fix anymore :) 

CleanShot 2022-10-28 at 12 53 49

alacuku:
@santrancisco, good catch! Anyway, I would definitely merge this PR for the sake of uniformity across the `values.yaml` file.

PR #453, opened 2023-01-20
new(falco): chart v3.0.0 for Falco 0.34 (falcoctl support and more)

### LINKED ISSUE #418: [driver-loader]: simplifying the logic behind the driver-loader
**Motivation**
The `driver-loader` is an init-container that is deployed by default when a driver (module/ebpf) is used. Users can override the default configuration and disable the `driver-loader`. There are two config variables used to disable the driver loader:
https://github.com/falcosecurity/charts/blob/06225b85abae5beb77e1a11cf0d393a8b1bb8ce8/falco/values.yaml#L188-L193
The two variables are used to configure the deployment such in this case:
https://github.com/falcosecurity/charts/blob/06225b85abae5beb77e1a11cf0d393a8b1bb8ce8/falco/templates/pod-template.tpl#L133-L146
It is not enough to just set the `driver.loader.enabled=false` in order to have the volumes mounted in the falco pod but we need to set the `driver.loader.initContainer.enabled=false`. 
I see that the `driver-loader` logic is decoupled from the `initContainer` since we can have the former `enabled` and the latter `disabled`. This allows the users to provide their own logic for the `driver-loader` but for the end user, it's confusing and error-prone. The helm chart will deploy Falco but it will not start. It's not straightforward to debug such cases.

**Feature**
I think we should simplify the `driver-loader` logic in our helm chart by setting just one configuration variable that enables or disables it. We should implement some `helpers` that mount the right volumes in the right containers based on the configuration. If we really want we can still allow people to provide their own logic for the driver loader by asking them to explicitly configure the `init containers`, `volumes`, and `volumes mounts`. 
We should aim to offer a functional and straightforward experience when deploying Falco by preventing unnecessary over-engineering of the charts.

### LINKED ISSUE #419: [falco/image]: falco errors when using "falcosecurity/falco" image
**Describe the bug**
The `falcosecurity` organizaton provides two different image flavours for Falco:
* falcosecurity/falco-driver-loader
* falcosecurity/falco
For more info: https://falco.org/docs/getting-started/download/#images.

Starting from version `2.0.0` of the charts the `falcosecurity/falco-no-driver` is the default image. As far as I know, we are still supporting the `falcosecurity/falco` image in our charts.

With the latest charts, v2.2.0, I get errors when using the `falcosecurity/falco` image.
 
In the first case, using the default values of the chart and just setting the `falcosecurity/falco` image:
```bash
helm install falco \
    --set image.repository=falcosecurity/falco \
    falcosecurity/falco
```
We have both the variables `driver.loader.enabled` and `driver.loader.initContainer.enabled` set to `true`.  The init container correctly downloads/builds the kernel module for our system but the `falco-driver-loader` script runs inside the Falco container and fails with:
```
* Looking for a Falco module locally (kernel 5.4.0-113-generic)
Detected an unsupported target system, please get in touch with the Falco community
```
That happens because we are not mounting the `/etc/` folder in the Falco container causing the script to fail. And it's ok because we are using the init container:
https://github.com/falcosecurity/charts/blob/06225b85abae5beb77e1a11cf0d393a8b1bb8ce8/falco/templates/pod-template.tpl#L133-L146

In the second case, we set to `false` both  `driver.loader.enabled` and `driver.loader.initContainer.enabled`:
```bash
helm install falco \
    --set image.repository=falcosecurity/falco \
    --set driver.loader.enabled=false \
    --set driver.loader.initContainer.enabled=false \
    falcosecurity/falco
```
The Falco pods does not start, and the error is:
```
  Warning  Failed     3s                kubelet            Error: failed to generate container "f41374f044c4a77e9a052ce42445fe1b0a962e3fbde6ae25e32b49e970734b04" spec: failed to generate spec: failed to mkdir "/sys/module/falco": mkdir /sys/module/falco: operation not permitted
```
In the latest version of the chart we add a new volume mount: `/sys/module/falco`. For more info why it is needed see here: https://github.com/falcosecurity/falco/pull/2214.
The `/sys/module/falco` folder is available only after the kernel module has been loaded.

In the third case, we set `driver.loader.enabled` to `true` but disable the `initContainer`:
```
helm install falco \
    --set image.repository=falcosecurity/falco \
    --set driver.loader.enabled=true \
    --set driver.loader.initContainer.enabled=false \
    falcosecurity/falco
```
We are removing the initContainer, but the logic of the `falco-driver-loader` script inside the Falco container should be triggered. We still get the error:
```
  Warning  Failed     4s                kubelet            Error: failed to generate container "f41374f044c4a77e9a052ce42445fe1b0a962e3fbde6ae25e32b49e970734b04" spec: failed to generate spec: failed to mkdir "/sys/module/falco": mkdir /sys/module/falco: operation not permitted
```

In order to resolve the mount issue without an init container we should mount the whole `/sys/module` folder inside the Falco container. And that is not a best practice since it is mounted in ReadWrite mode.

**How to reproduce it**
Follow the steps described above.
**Expected behaviour**
The Falco pods should be up and running after the deployment.

loresuso:
Signed-off-by: Aldo Lacuku <aldo@lacuku.eu>
Co-authored-by: Lorenzo Susini <susinilorenzo1@gmail.com>

**What type of PR is this?**

/kind bug

/kind cleanup

/kind design

/kind documentation

/kind feature

/kind chart-release

/area falco-chart

- [x] Add support for `falcoctl` init container and sidecar container. Opening this on WIP for early feedback.
- [x] Fixes #418 
- [x] Fixes #419  
- [x] Remove default rules files from the chart 
- [x] Add modern probe support (TBD see https://github.com/falcosecurity/charts/pull/459)
- [x] Upgrade to Falco 0.34

Fixes #418 
Fixes #419

**Checklist**

- [x] Chart Version bumped
- [x] CHANGELOG.md updated


falco/templates/_helpers.tpl
```diff
@@ -229,3 +229,70 @@ be temporary and will stay here until we move this logic to the falcoctl tool.
...
+- name: falcoctl-artifact-install
+  image: {{ printf "%s/%s:%s" .Values.falcoctl.image.registry .Values.falcoctl.image.repository .Values.falcoctl.image.tag }}
+  imagePullPolicy: {{ .Values.falcoctl.image.pullPolicy }}
+  args: 
+    {{- $str := "[\"artifact\", \"install\", " }}
+    {{- range $key, $value := .Values.falcoctl.artifact.install.args }}
+    {{- $str = print $str "\"" $value "\"" "," }}
+    {{- end }}
+    {{- $str = print $str "]" }}
+    {{- $str | nindent 4 }}
```
loresuso, line 244: This is the way I found to concatenate args. Don't know if there is something smarter that can be done here!

leogr: This :point_down: should be more straightforward! :angel: 

https://github.com/falcosecurity/charts/blob/9b27a5925f4c8a4d0fd500a2e963d572e2cce767/falco/templates/pod-template.tpl#L93-L95

falco/templates/_helpers.tpl
```diff
@@ -229,3 +229,70 @@ be temporary and will stay here until we move this logic to the falcoctl tool.
     - mountPath: /gvisor-config
       name: falco-gvisor-config
 {{- end -}}
+
+
+{{- define "falcoctl.initContainer" -}}
+- name: falcoctl-artifact-install
+  image: {{ printf "%s/%s:%s" .Values.falcoctl.image.registry .Values.falcoctl.image.repository .Values.falcoctl.image.tag }}
```
leogr, line 236: I think it's better to create a helper here. We already do that for the other images. For example :point_down: 

https://github.com/falcosecurity/charts/blob/9b27a5925f4c8a4d0fd500a2e963d572e2cce767/falco/templates/pod-template.tpl#L55

falco/templates/_helpers.tpl
```diff
@@ -229,3 +229,70 @@ be temporary and will stay here until we move this logic to the falcoctl tool.
...
+  env:
+  {{- range $key, $value := .Values.falcoctl.artifact.install.env }}
+    - name: "{{ $key }}"
+      value: "{{ $value }}"
+  {{- end }}
+{{- end -}}
+
+{{- define "falcoctl.sidecar" -}}
+- name: falcoctl-artifact-follow
+  image: {{ printf "%s/%s:%s" .Values.falcoctl.image.registry .Values.falcoctl.image.repository .Values.falcoctl.image.tag }}
```
leogr, line 269: Same here.

falco/templates/_helpers.tpl
```diff
@@ -229,3 +229,70 @@ be temporary and will stay here until we move this logic to the falcoctl tool.
...
+  imagePullPolicy: {{ .Values.falcoctl.image.pullPolicy }}
+  args:
+    {{- $str := "[\"artifact\", \"follow\", " }}
+    {{- range $key, $value := .Values.falcoctl.artifact.install.args }}
+    {{- $str = print $str "\"" $value "\"" "," }}
+    {{- end }}
+    {{- $str = print $str "]" }}
+    {{- $str | nindent 4 }}
```
leogr, line 277: Same here.

falco/templates/pod-template.tpl
```diff
@@ -134,6 +134,10 @@ spec:
           {{- end }}
       {{- end }}
       volumeMounts:
+        - mountPath: /etc/falco/falcoctl-rules
```
leogr, line 137: This differs from the default Falco configuration. How do we enforce this in the Falco config?

falco/values.yaml
```diff
@@ -12,7 +12,7 @@ image:
   # -- The image repository to pull from
   repository: falcosecurity/falco-no-driver
   # -- The image tag to pull. Overrides the image tag whose default is the chart appVersion.
-  tag: ""
+  tag: "master"
```
leogr, line 15:
```suggestion
  tag: ""
```

Please don't forget to revert this.

falco/values.yaml
```diff
@@ -348,6 +399,8 @@ falco:
     - /etc/falco/falco_rules.yaml
     - /etc/falco/falco_rules.local.yaml
     - /etc/falco/rules.d
+    - /etc/falco/falcoctl-rules
+
```
leogr, line 403: This should not work as expected because the `falco_rules.yaml` will be downloaded in another folder, and as result, Falco will crash (rules will be loaded twice).

Moreover, this approach is fragile (see one of the comments above, where I wonder how we can enforce `/etc/falco/falcoctl-rules`).

That being said, I don't have a solution yet :sweat_smile:

leogr (commented):
Great job @loresuso ! :star_struck: 

This PR goes almost in the correct direction, there're only a couple of things to be improved, and just an issue to be addressed.

See my comments.

falco/values.yaml
```diff
@@ -327,6 +327,57 @@ falcosidekick:
...
+      # -- Security context for the falcoctl init container.
+      securityContext: {}
+  
+  config:
+    indexes:
+    - name: falcosecurity
+      url: https://falcosecurity.github.io/falcoctl/index.yaml
+    artifact:
+      install:
+        refs: [ falco-rules:0.1 ]
```
leogr, line 371:
```suggestion
        refs: [ falco-rules:0 ]
```

falco/values.yaml
leogr: Side note: order matters here. To improve the UX, we should also pay attention to loading custom user-provided rules, if any, always _after_ those already installed.

loresuso:
@leogr we addressed all the comments. Could you please check again? Thanks!

falco/values.yaml
```diff
@@ -203,7 +203,7 @@ driver:
         # -- The image repository to pull from.
         repository: falcosecurity/falco-driver-loader
         #  -- Overrides the image tag whose default is the chart appVersion.
-        tag: ""
+        tag: "master"
```
leogr, line 206:
```suggestion
        tag: ""
```
Same as above

falco/values.yaml
```diff
@@ -384,7 +456,10 @@ falco:
   # load_plugins: [cloudtrail, json]
   # -- Add here the names of the plugins that you want to be loaded by Falco. Please make sure that
   # plugins have been configured under the "plugins" section before adding them here.
-  load_plugins: []
+  # Please make sure to configure the falcoctl tool to download and install the very same plugins
+  # you are loading here. You should add the references in the falcoctl.config.artifact.install.refs array
+  # for each plugin you are loading.
+  load_plugins: [k8saudit, json]
```
leogr, line 462:
```suggestion
  load_plugins: []
```
Revert this pls.

falco/values.yaml
```diff
@@ -327,6 +327,78 @@ falcosidekick:
...
+    artifact:
+      install:
+        # -- List of artifacts to be installed by the falcoctl init container.
+        refs: [falco-rules:0]
+        # -- Directory where the rulesfiles are saved. The path is relative to the container, which in this case is an emptyDir
+        # mounted also by the Falco pod. If you change this value, please make sure you change the mount path of the emptyDir inside
+        # the falcoctl-artifact-install init container.
+        rulesfilesdir: /rulesfiles
+        # -- Same as the one above but for the artifacts.
+        pluginsdir: /plugins
```
leogr, line 388: This is not completely clear to me.

How could the user `change the mount path of the emptyDir inside...` since those values are hard-coded in the chart :point_down: ?

https://github.com/falcosecurity/charts/pull/453/files#diff-701550ac61b2456be7b7355b821a218487fff642078f5d0aa22f1be38c725a75R260-R263

We may use those values to actually set the mount path inside the container OR disallow the user to change those values.
I prefer the first option (IIRC, we already to the same with some other option, maybe the gRPC socket path? :thinking:). However, I haven't thoughts if there're implications.

falco/templates/_helpers.tpl
```diff
@@ -229,3 +236,68 @@ be temporary and will stay here until we move this logic to the falcoctl tool.
...
+  {{- end }}
+  securityContext:
+  {{- if .Values.falcoctl.artifact.install.securityContext }}
+    {{- toYaml .Values.falcoctl.artifact.install.securityContext | nindent 4 }}
+  {{- end }}
+  volumeMounts:
+    - mountPath: /plugins
+      name: plugins-install-dir
+    - mountPath: /rulesfiles
+      name: rulesfiles-install-dir
```
leogr, line 263: Probably, `/plugins` and `/rulesfiles` shouldn't be hardcoded.
See my comment below (the one attached to the config).

falco/templates/pod-template.tpl
```diff
@@ -134,6 +134,10 @@ spec:
           {{- end }}
       {{- end }}
       volumeMounts:
+        - mountPath: /etc/falco
+          name: rulesfiles-install-dir
+        - mountPath: /usr/share/falco/plugins
+          name: plugins-install-dir
```
leogr, line 140: Unless I'm missing something...
Should those paths be mounted only when falcoctl's install or follow features are enabled? :thinking:

leogr (commented):
This is impressive! :star_struck: 

I've left some comments, but overall this PR looks great.

Also, please bump the chart version and add a bit of documentation :pray: 
I propose bumping the chart to 3.0.0 since the default behavior is significatively changed.

falco/templates/NOTES.txt
```diff
@@ -23,3 +23,9 @@ Full list of outputs: https://github.com/falcosecurity/charts/tree/master/falcos
 You can enable its deployment with `--set falcosidekick.enabled=true` or in your values.yaml. 
 See: https://github.com/falcosecurity/charts/blob/master/falcosidekick/values.yaml for configuration values.
 {{- end}}
+
+{{- if and (not (empty .Values.falco.load_plugins)) (or .Values.falcoctl.artifact.follow.enabled .Values.falcoctl.artifact.install.enabled) }}
+
+WARNING:
+{{ printf "It seems you are loading the following plugins %v, please make sure to install them by adding the correct reference to falcoctl.config.artifact.install.refs: %v" .Values.falco.load_plugins .Values.falcoctl.config.artifact.install.refs -}}
+{{- end }}
```
leogr, line 31: Not sure about this. k8saudit and cloudtrail are included in the container image, and it shouldn't be a hard requirement to use falcoctl to use them.

That should be working already since when both the _install_ and _follow_ are disabled, no empty dir is mounted.

falco/templates/pod-template.tpl
```diff
@@ -98,10 +98,6 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: spec.nodeName
-      {{- if or (not .Values.driver.enabled) (and .Values.driver.loader.enabled .Values.driver.loader.initContainer.enabled) }}
-        - name: SKIP_DRIVER_LOADER
-          value:
-      {{- end }}
```
leogr, line 104: Please, that should be documented in the changelog.

falco/values-k8saudit.yaml
```diff
@@ -1,11 +1,34 @@
...
+
+
+falcoctl:
+  artifact:
+    install:
+      # -- Enable the falcoctl tool as init container. It installs artifacts in the config.artifact.install.refs list.
+      enabled: true
+    follow:
+      # -- Disable the sidecar container. We do not support it yet for plugins. It is used only for rules feed such as Falco rules.
+      enabled: false
```
leogr, line 25: We may want to enable this only for `k8saudit-rules`. Perhaps, just explaining that in the comment would be enough for now.

falco/values-syscall-k8saudit.yaml
```diff
@@ -0,0 +1,64 @@
...
+    follow:
+      enabled: true
+  config:
+    artifact:
+      install:
+        # -- List of artifacts to be installed by the falcoctl init container.
+        # Note that the plugins that are loaded by falco are set here. We need to download and install whem using
+        # the falcoctl init container. Once installed they are available to Falco under /usr/share/falco/plugins folder. 
+        # See "load_plugins" section.
+        refs: [falco-rules:0, k8saudit:0, json:0]
```
leogr, line 34:
```suggestion
        refs: [falco-rules:0, k8saudit-rules:0, k8saudit:0, json:0]
```

falco/values-k8saudit.yaml
```diff
@@ -1,11 +1,34 @@
...
+  config:
+    artifact:
+      install:
+        # -- List of artifacts to be installed by the falcoctl init container.
+        # Same plugins we are loading in Falco. See "load_plugins" section.
+        refs: [k8saudit:0, json:0]
```
leogr, line 31:
```suggestion
        refs: [k8saudit-rules:0, k8saudit:0, json:0]
```

falco/values-syscall-k8saudit.yaml
```diff
@@ -0,0 +1,64 @@
...
+      follow:
+        # -- List of artifacts to be followed by the falcoctl sidecar container.
+        refs: [falco-rules:0]
```
leogr, line 37: Perhaps we want to add `k8saudit-rules:0` too.

falco/values.yaml
```diff
@@ -329,6 +327,77 @@ falcosidekick:
...
+    tag: "main"
+  artifact:
+    # -- Runs "falcoctl artifact install" command as an init container. It is used to install artfacts before
+    # Falco starts. It provides them to Falco by using an emptyDir volume.
+    install:
+      enabled: true
+      # -- Extra environment variables that will be pass onto falcoctl-artifact-install init container.
+      env: {}
+      # -- Arguments to pass to the falcoctl-artifact-install init container.
+      args: ["--verbose"]
```
leogr, line 351: Why do we want `--verbose` by default?

falco/values.yaml
```diff
@@ -329,6 +327,77 @@ falcosidekick:
...
+    # updates given a list of artifacts. If an update is found it downloads and installs it in a shared folder (emptyDir)
+    # that is accessible by Falco. Rulesfiles are automatically detected and loaded by Falco once they are installed in the
+    # correct folder by falcoctl. To prevent new versions of artifacts from breaking Falco, the tool checks if it is compatible
+    # with the running version of Falco before installing it.
+    follow:
+      enabled: true
+      # -- Extra environment variables that will be pass onto falcoctl-artifact-follow sidecar container.
+      env: {}
+      # -- Arguments to pass to the falcoctl-artifact-follow sidecar container.
+      args: ["--verbose"]
```
leogr, line 366: Why do we want `--verbose` by default?

leogr (commented):
Another review iteration and this PR is getting better!

I propose to retitle this PR and use it for all changes related to Falco 0.34.

Also, it misses the following:
- to bump the chart version (it's now clear to me we should set `3.0.0`
- update the changelog
- update the readme

Anyway, great job :clap:

falco/templates/NOTES.txt
alacuku: The warning is printed only if we are loading plugins and one of `falcoctl.artifact` or `falcoctl.install`. So yes, it does what you are saying.

falco/values.yaml
alacuku: I don't have a strong opinion on this.

falco/templates/NOTES.txt
leogr: You are right!

falco/values-k8saudit.yaml
leogr, line 31:
```suggestion
falcoctl:
  artifact:
    install:
      # -- Disable the init container. We do not recommend installing (or following) plugins for security reasons since they are executable objects.
      enabled: false
    follow:
      # -- Enable the sidecar container. We do not support it yet for plugins. It is used only for rules feed such as k8saudit-rules rules.
      enabled: true
  config:
    artifact:
      follow:
        # -- List of artifacts to be followed by the falcoctl sidecar.
        # Assuming plugins k8saudit v0.5.0 and json v0.6.0 shipped in-bundle with Falco 0.34
        refs: [k8saudit-rules:0.5]
```

leogr (commented):
As discussed privately with @alacuku and @loresuso, the only missing thing here is a way to disallow an unwanted plugin installation (since it's not yet fully supported in our installation scenario and security concerns may apply).

Potential solution: https://github.com/falcosecurity/falcoctl/pull/251

Otherwise SMGT.

cc @LucaGuerra (since you are the release manager for Falco 0.34), 

I think it is worth delaying a release by 1 day if we are able to get this, but let's decide together.

PS

Thank you'll, this is an amazing job :pray:

leogr (dismissed):
:+1: 

/hold 
feel free to remove the "hold" when needed
cc @loresuso @alacuku @LucaGuerra

loresuso:
We also fixed links in the docs. Mind to reapprove @leogr? :)

leogr (approved):

leogr:
/close

Closing and reopening to trigger the CI

PR #480, opened 2023-04-12
fix: use SYS_ADMIN and SYS_RESOURCE capabilities since SYS_PERFMON is not sufficient

jemag:
Related to falcosecurity/falco#2487

**What type of PR is this?**

/kind bug
/kind chart-release

/area falco-chart

SYS_PERFMON capability is not sufficient to run falco with `leastPrivileged: true`, see falcosecurity/falco#2487 for details

I am not sure if falcosecurity/falco#2487 would be fixed since I believe some investigation will be done on why this is the case with SYS_PERFMON, but this makes the chart usable with  `leastPrivileged: true`

**Checklist**
- [x ] Chart Version bumped
- [x ] CHANGELOG.md updated

### DIFF
--- a/falco/CHANGELOG.md
+++ b/falco/CHANGELOG.md
@@ -3,6 +3,9 @@
 This file documents all notable changes to Falco Helm Chart. The release
 numbering uses [semantic versioning](http://semver.org).
 
+## v3.1.4
+
+* Use `SYS_ADMIN` and `SYS_RESOURCE` capabilities since `SYS_PERFMON` is not sufficient
 ## v3.1.3
 
 * Updates the grpc-service to use the correct label selector
--- a/falco/Chart.yaml
+++ b/falco/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: falco
-version: 3.1.3
+version: 3.1.4
 appVersion: 0.34.1
 description: Falco
 keywords:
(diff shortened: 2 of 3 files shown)


Andreagit97:
Hey @jemag thank you for the fix! before merging it we would like to investigate more on how to proceed. The old bpf probe requires `SYS_ADMIN` but the modern one is able to use just `PERFMON` and `BPF` so probably we will have a different security context according to the driver we use...but as i said we need to investigate a little bit deeper :)

jemag:
I thought the `modern-bpf` wasn't able to run without privileged, given that it seems to run with `privileged: true` in the helm chart:
https://github.com/falcosecurity/charts/blob/6fff4e1e75c43af742596f478bf86e7723aa08da/falco/templates/pod-template.tpl#L392-L402
and
https://github.com/falcosecurity/charts/blob/6fff4e1e75c43af742596f478bf86e7723aa08da/falco/values.yaml#L69-L76

I think it is definitely great to investigate it further. That being said, if you think the investigation required could take weeks to months, then it might still be worth it to provide end-users of the helm chart a working  `leastPrivileged: true` in the meantime (even if this means changing this again down the line or reverting).

alacuku:
Hi @jemag, #506 fixes the `least privileged` mode for the modern bpf.

jemag:
@alacuku Thanks, been pretty busy but will definitely  take a look soon

PR #482, opened 2023-04-19
fix falco host path mount of /sys/module

Kaizhe:
**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /kind chart-release

The current setting blocks my Falco 0.34.1 deployment in EKS, I manually modified the mounting path to `/sys/module` and it works.

 /area falco-chart

> /area falco-exporter-chart

> /area falcosidekick-chart

> /area event-generator-chart

Fixes #

Falco was mounted on directory `/sys/module/falco`, since the directory doesn't exist. Kubelet need to create the directory `falco` under `/sys/module`. However, `/sys/module` is a virtual filesystem, normal directory creation doesn't work in that directory, even with root user.  Hence, mount on `/sys/module` instead of `/sys/module/falco`.

The current setting blocks my falco 0.34.1 deployment in EKS. After I modified the host path mount to `/sys/module`, it works.

**Checklist**

- [x] Chart Version bumped
- [x] CHANGELOG.md updated

### DIFF
--- a/falco/CHANGELOG.md
+++ b/falco/CHANGELOG.md
@@ -3,6 +3,10 @@
 This file documents all notable changes to Falco Helm Chart. The release
 numbering uses [semantic versioning](http://semver.org).
 
+## v3.1.4
+
+* Mount on `/sys/module` instead of `/sys/module/falco` for kernel module version.
+
 ## v3.1.3
 
 * Updates the grpc-service to use the correct label selector
--- a/falco/Chart.yaml
+++ b/falco/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: falco
-version: 3.1.3
+version: 3.1.4
 appVersion: 0.34.1
 description: Falco
 keywords:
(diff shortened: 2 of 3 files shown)


alacuku:
The `/sys/module/falco` is needed by Falco to configure kernel buffers, and we are mounting only the `falco` folder. We do not want to mount the whole `/sys/module` in the Falco container. 

The `/sys/module/falco` folder is available once the `kernel module` is loaded, so we assume that Falco container is started only if the `falco-driver-loader` init container has successfully loaded the `kernel module`. Even in the case when you are not using the `falco-driver-loader` init container you would need to load the `kernel module` by other means. So if the kubelet is refusing to start the Falco pod than it means that the `kernel module` is not loaded in the system. The patch is just postponing the failure of the Falco pod. Once it is up, it will attempt to start the capture using the kernel module, but will fail since the `kernel module` is not loaded.

PR #528, opened 2023-09-18
Revert "update falco chart to 3.6.1"

Issif:
Reverts falcosecurity/charts#527


cpanato:
you need to move forward never backwards

PR #536, opened 2023-09-25
:sparkles: Allow annotations from falco's dependency

afreyermuth98:
**What type of PR is this?**

/kind design

/kind chart-release

/area falco-chart

This PR allows to pass some annotations to the falcosidekick dependency. It's useful as for example I want to scrape this service using prometheus annotations.

Fixes #

**Checklist**

- [x] Chart Version bumped
- [x] CHANGELOG.md updated

### DIFF
--- a/falco/values.yaml
+++ b/falco/values.yaml
@@ -338,6 +338,11 @@ falcosidekick:
   fullfqdn: false
   # -- Listen port. Default value: 2801
   listenPort: ""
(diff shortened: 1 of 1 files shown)


Issif:
Hi,

These values are already present in the falcosidekick's values.yaml, no need to repeat them there. The docs explicitly says, to configure falcosidekick itself, you need to check the existing values and add the prefix `falcosidekick.`.

PR #546, opened 2023-10-13
Feature: Deploy falcosidekick and falco server and client secrets for mTLS

jgmartinez:
**What type of PR is this?**

/kind feature

/kind chart-release

/area falco-chart

/area falcosidekick-chart

This PR adds the capability of loading certificates dynamically via helm values, instead mounting volumes. It's structured in a way to make it easier to deploy mTLS cryptographic material for both falco and falcosidekick when http_output is enabled.

N/A

There are some changes in the directories to store the certificates. It seems not to break anything, but this of course can be overridden by a local values file. 

**Checklist**

- [X] Variables are documented in the README.md

### DIFF
--- a/falco/templates/client-certs-secret.yaml
+++ b/falco/templates/client-certs-secret.yaml
@@ -1,0 +1,21 @@
+{{- if and .Values.certs.client.key .Values.certs.client.crt .Values.certs.ca.crt }}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ include "falco.fullname" . }}-client-certs
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app.kubernetes.io/name: {{ include "falco.name" . }}
+    helm.sh/chart: {{ include "falco.chart" . }}
+    app.kubernetes.io/instance: {{ .Release.Name }}
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
+type: Opaque
+data:
+  {{ $key := .Values.certs.client.key }}
+  client.key: {{ $key | b64enc | quote }}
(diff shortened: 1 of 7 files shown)


Issif:
By changing the folders for the certs, are you sure it will not create breaking change?

Moreover I'm not sure the CI will accept to merge a PR for modifications of 2 charts at once.

jgmartinez:
Hi @Issif! 

You are right, changing the default values of the cert directory, in the Falco chart, could break https connections if no CA is defined. I think it would be better to leave it blank since Falco itself defaults to `/etc/ssl/certs` when no value is defined for ca_cert in this [line](https://github.com/falcosecurity/falco/blob/e4054fc5ea14eb0872df2e583a87c532f7504ec3/userspace/falco/configuration.cpp#L211). Same for [client_cert](https://github.com/falcosecurity/falco/blob/e4054fc5ea14eb0872df2e583a87c532f7504ec3/userspace/falco/configuration.cpp#L220) and [client_key](https://github.com/falcosecurity/falco/blob/e4054fc5ea14eb0872df2e583a87c532f7504ec3/userspace/falco/configuration.cpp#L223C6-L223C6).

As of Falcosidekick, I just added the `falcosidekick` directory name for clarity, similarly as in [falco-exporter](https://github.com/falcosecurity/charts/blob/master/falco-exporter/templates/daemonset.yaml#L88), I can rollback to the previous name if it's preferred. 

Confirm if I need to open separate PRs for each chart and I'll do it when all changes are "approved" in a comment 😄 

Thank you!

jgmartinez:
Split to #548 and #549 @Issif

PR #595, opened 2023-12-22
[WIP] [Falco 0.37.0] cleanup(falco): remove `--userspace` flag

Andreagit97:
**What type of PR is this?**

/kind cleanup

/area falco-chart

This PR removes the `--userspace` flag from Falco helm chart

**Checklist**

- [ ] Chart Version bumped
- [ ] CHANGELOG.md updated

### DIFF
--- a/charts/falco/ci/ci-values.yaml
+++ b/charts/falco/ci/ci-values.yaml
@@ -6,11 +6,10 @@
 driver:
   enabled: false
 
-# make Falco run in userspace only mode
-extra:
-  args:
-    - --userspace
-
+falco:
+  # Run Falco in no driver mode, to avoid injection of kernel modules or eBPF probes
+  engine:
+    kind: none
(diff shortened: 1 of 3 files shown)


Andreagit97:
This PR is just to remind us that we need to remove the `--userspace` flag before releasing the helm chart, because it doesn't exist anymore. This PR will fail until we bump the Falco version in the helm chart since the new `engine.kind` config is supported only in Falco 0.37.0. 

Some side notes:
* do we still need the `--disable-source` flag, now that we have `kind.engine = none`
* do we really need the block `driver` in the helm chart `.yaml`? why not use the new Falco config?

Andreagit97:
> do we really need the block driver in the helm chart .yaml? why not use the new Falco config?

In the end yes we need to keep it, because there are `leastPrivileged` and `hostNetwork` fields that are only inherent to the deployment of Falco

Andreagit97:
done in #601

PR #710, opened 2024-07-11
Fix redis-availability check in UI

jegger:
**What type of PR is this?**

/kind bug
/kind chart-release

/area falco-chart
/area falcosidekick-chart

It fixes the redis-availability check of the UI init-container in case external-redis is enabled. It was using the hard-coded value used when you would not use an exernal-redis.

I also upgraded the chart-version of the sidekick and flaco to get the fix also into the falco chart. It would certainly support the use I have - but I don't know if bumping two versions at once is something you do. Let me know.

- [x] Chart Version bumped
- [x] CHANGELOG.md updated


Issif (commented):
The changes are good, but you can mix in the same PR the update  the Falcosidekick and Falco charts. First update Falcosidekick and once merge, update Falco. This is required by the CI. Thanks.

Issif (approved):

PR #779, opened 2024-11-11
fix(falcosidekick): Allow overriding the Prometheus scrape Service annotation

TiagoJMartins:
**What type of PR is this?**

/kind bug
/kind chart-release

/area falcosidekick-chart

We are working on a migration from a deprecated Prometheus setup to one based on the Prometheus Kubernetes Operator.
Since metrics scraped by the operator-managed instances are targeting ServiceMonitor and PodMonitor resources, we need to be able to configure the `prometheus.io/scrape` Service annotation to gradually disable scraping from our deprecated instances.

Because the `prometheus.io/scrape` annotation is hard-coded in the Service template **and** it's the last entry in the `annotations` map, there's no way to include custom values after it.

This PR simply moves the hard-coded annotation to before the user-provided values, so at least they have an escape hatch in case they need to override the value.

You can test with:
```sh
helm template falcosidekick charts/falcosidekick -s templates/service.yaml -f - <<EOF | yq '.metadata.annotations["prometheus.io/scrape"]'
service:
  annotations:
    prometheus.io/scrape: "false"
EOF
```

The final template will include duplicate values, but the override will take precedence when the yaml gets parsed/applied to the cluster.
```yaml
apiVersion: v1
kind: Service
metadata:
  name: falcosidekick
  annotations:
    prometheus.io/scrape: "true"
    prometheus.io/scrape: "false"
  ...
```

The duplicated keys will be removed by the cluster.

I considered going with a different approach based on something like `.Values.service.prometheusScrape`, but decided against it due to it requiring changes to the `values.yaml` structure.

**Checklist**

- [x] Chart Version bumped
- [x] CHANGELOG.md updated

### DIFF
--- a/charts/falcosidekick/CHANGELOG.md
+++ b/charts/falcosidekick/CHANGELOG.md
@@ -5,6 +5,10 @@ numbering uses [semantic versioning](http://semver.org).
 
 Before release 0.1.20, the helm chart can be found in `falcosidekick` [repository](https://github.com/falcosecurity/falcosidekick/tree/master/deploy/helm/falcosidekick).
 
+## 0.9.5
+
+- Move the `prometheus.io/scrape` annotation to the default values, to allow overrides.
+
 ## 0.9.4
 
 - Fix Prometheus metrics names in Prometheus Rule
--- a/charts/falcosidekick/Chart.yaml
+++ b/charts/falcosidekick/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: 2.30.0
 description: Connect Falco to your ecosystem
 icon: https://raw.githubusercontent.com/falcosecurity/falcosidekick/master/imgs/falcosidekick_color.png
 name: falcosidekick
-version: 0.9.4
+version: 0.9.5
 keywords:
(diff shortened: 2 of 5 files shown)


rsicart:
And what about removing the hard-coded annotation and putting it by default in `.Values.service.annotations`?

Issif:
Can you rebase on master and fix the conflict please? 

> And what about removing the hard-coded annotation and putting it by default in `.Values.service.annotations`?

I think this comment is valuable, wdyt?

TiagoJMartins:
> Can you rebase on master and fix the conflict please?
> 
> > And what about removing the hard-coded annotation and putting it by default in `.Values.service.annotations`?
> 
> I think this comment is valuable, wdyt?

Sorry for the delay; branch updated!

Issif:
can you run `make docs-falcosidekick`, the CI complains about a missing change in the readme because of that. thanks

leogr:
Hey @Issif 

Are we ok with this now?

Issif (approved):

Issue #825, opened 2025-01-31
caPEMBundle missing causes problems for default installs.


data-dude:
When falco talks to the k8s-metacollector by default ssl isn't setup.  But if caPEMBundle isn't specifically set to spaces in the k8smta configuration you will run into ssl issues.   If you don't include the caPEMBundle configuration it seems to assume you do what ssl.     This causes issues for initial installs. 
k8smeta 0.2.0
chart 4.17.2
The easiest fix is to set the caPEMBundle to spaces in the helm chart.

alacuku:
Hey @data-dude, can you share some more info on how are you installing falco? what is your `values.yaml` file? The default configuration does not set the values for the certificates, and it works fine. I think the issue you are facing is related to your enviroment/installation tools.

data-dude:
This doesn't appear to be an issue

PR #835, opened 2025-02-05
capembundle needs to be set to avoid ssl errors

data-dude:
This updates falco/values.yaml and falco/templates/_helper
If I didn't set caPEMBundle to spaces I ran into ssl errors.  k8s-metacollector by default doesn't use ssl but if you don't set caPEMBundle to spaces k8smeta plugin seems to assume you are using ssl.
This change allows users to set it to spaces and sets it to that by default.

> /kind bug

> /area falco-chart

https://github.com/falcosecurity/charts/issues/825


data-dude:
This doesn't appear to be an issue

PR #899, opened 2025-08-06
Removed allowedUnsafeSysctls from SCC

### LINKED ISSUE #898: allowedUnsafeSysctls cannot be set to an empty array in SecurityContextConstraint
**Describe the bug**

The allowedUnsafeSysctls field in a SecurityContextConstraint (SCC) defines which sysctl calls a workload (such as Falco) can make in OpenShift/OKD.
In this case, the expected empty value is none, but it was set to []. This causes the OpenShift API server to drop the field entirely.

While an SCC without this field still works fine for Falco, using [] can cause synchronization issues with tools like ArgoCD, because the SCC in the cluster will never match the manifest’s expected allowedUnsafeSysctls: [].

Solution: Remove the allowedUnsafeSysctls field entirely — it’s not needed.

More info: [Red Hat Documentation – allowedUnsafeSysctls](https://docs.redhat.com/en/documentation/openshift_container_platform/4.18/html/security_apis/securitycontextconstraints-security-openshift-io-v1#specification-8)

**How to reproduce it**

Deploy falco helm chart in OKD/Openshift using ArgoCD or similar tools.

**Expected behaviour**

Falco helm chart should not set fields to values that are not allowed and rely on the openshift api server to revert it.

**Screenshots**

Image

**Environment**

- Falco version: any version of the chart is affected
- System info: any Openshift/OKD version is affected)
- Cloud provider or hardware configuration: not related
- OS: not related
- Kernel: not related
- Installation method: Kubernetes (Openshift)

AlbertoPimpo:
**What type of PR is this?**

/kind bug
/kind chart-release

/area falco-chart

One liner fix for #898, more context in the issue.

Fixes #898 

/cc @airadier 

**Checklist**
- [ ] Chart Version bumped
- [ ] CHANGELOG.md updated

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,8 +16,8 @@ lint-%:
 	-e HOME=/home/ct \
 	--mount type=tmpfs,destination=/home/ct \
 	--workdir=/data \
-	--volume $$(pwd):/data \
-	-u $$(id -u) \
+	--volume $$(pwd):/data:Z \
+	--userns=host \
 	quay.io/helmpack/chart-testing:$(LINT_IMAGE_VERSION) \
 	ct lint --config ./ct.yaml --charts ./charts/$*
 
@@ -28,8 +28,8 @@ docs-%:
 	@docker run \
 	--rm \
 	--workdir=/helm-docs \
-	--volume "$$(pwd):/helm-docs" \
-	-u $$(id -u) \
+	--volume "$$(pwd):/helm-docs:Z" \
+	--userns=host \
 	jnorwood/helm-docs:$(DOCS_IMAGE_VERSION) \
 	helm-docs -c ./charts/$* -t ./README.gotmpl -o ./README.md
 


leogr:
Hey @AlbertoPimpo 

I replied to you on this [slack thread](https://kubernetes.slack.com/archives/CMWH3EH32/p1755002388640019?thread_ts=1754471126.266079&cid=CMWH3EH32). May you bump the version, update the changelog, and then run `make docs-falco` (from the project root) please?

:pray:

leogr:
> I just added the `:Z` parameter to propagate selinux labels to the container (if selinux is not on use, nothing will change) and using ` --userns=host ` instead of setting manually the userid (which for some reasons is not working in podman).
> 
> Shall I create a new MR with this change?

Since the change is really small, feel free to include it in this PR directly!

And thanks for fixing this :heart:

leogr (approved):

PR #991, opened 2026-04-15
fix(falco): make FALCO_HOSTNAME env var conditional and suppressible

dunkin0486:
/kind feature
/kind chart-release
/area falco-chart

Makes the chart-default `FALCO_HOSTNAME` env var conditional and suppressible via a new `falco.hostnameEnv` values key (default `true`, backward-compatible), plus auto-detection of user-provided overrides in `extra.env`.

The chart currently injects `FALCO_HOSTNAME` from `spec.nodeName` unconditionally:

https://github.com/falcosecurity/charts/blob/master/charts/falco/templates/pod-template.tpl#L85-L88

Users with hostname collisions against other agents (Datadog, etc.) want to override `FALCO_HOSTNAME` to a custom literal. The natural approach is `extra.env`, but Kubernetes uses `env[].name` as the strategic-merge-patch merge key. Two entries with the same name collapse on UPDATE into one entry retaining fields from both -- yielding `value` + `valueFrom` on the same EnvVar, which fails validation:

```
EnvVar.valueFrom: Invalid value: "": may not be specified when value is not empty
```

This is unfixable at the K8s layer (kubernetes/kubernetes#44111, closed wontfix in 2017). Downstream users currently work around it by forking the chart to delete the default. A small chart-side conditional removes the need for forks.

**How**

Three modes are now supported:

| Mode | How user expresses it | Result |
|------|----------------------|--------|
| Default (status quo) | leave `falco.hostnameEnv` unset; no `FALCO_HOSTNAME` in `extra.env` | chart-default `valueFrom: spec.nodeName` injected (unchanged behavior) |
| Override custom value | put `FALCO_HOSTNAME` in `extra.env` | chart-default suppressed; user's value wins; no duplicate |
| Suppress entirely | set `falco.hostnameEnv: false` | no `FALCO_HOSTNAME` env var emitted at all |

Fixes (no open issue against the chart yet -- related: falcosecurity/falco#3397)

* Backward-compatible: default behavior is unchanged.
* `FALCO_K8S_NODE_NAME` immediately below has the same `valueFrom` pattern and would hit the same K8s limitation if anyone tried to override it. Kept out of scope here to keep this PR small; happy to file a follow-up if the maintainers want symmetric treatment.
* Tested locally with \`helm template\` for all three modes; \`make lint\` clean.

**Checklist**

- [x] Chart Version bumped (8.0.2 -> 8.0.3)
- [x] Variables are documented in the README.md (regenerated via `make docs`)
- [x] CHANGELOG.md updated


charts/falco/values.yaml
```diff
@@ -708,6 +708,19 @@ grafana:
...
+  ##################################
+
+  # -- Inject the chart-default `FALCO_HOSTNAME` env var (sourced from
+  # `spec.nodeName` via `valueFrom.fieldRef`) into the Falco container.
+  # Set to `false` to suppress entirely. Alternatively, override the value by
+  # adding a `FALCO_HOSTNAME` entry to `extra.env` -- the chart auto-detects
+  # this and suppresses the default, avoiding a Kubernetes strategic-merge-patch
+  # collision on `env[].name` during UPDATE that otherwise produces an invalid
+  # EnvVar with both `value` and `valueFrom` set.
+  hostnameEnv: true
```
c2ndev: I would avoid putting this under `falco:`. That object is rendered into `falco.yaml`, so `hostnameEnv` becomes Falco runtime config.

What about moving it to the chart-level pod/env settings?

```yaml
# -- Inject the chart-default `FALCO_HOSTNAME` env var.
# If `extra.env` contains `FALCO_HOSTNAME`, the user-provided value wins.
falcoHostnameEnv: true
```

charts/falco/templates/pod-template.tpl
```diff
@@ -82,10 +82,24 @@ spec:
...
+          {{- if eq .name "FALCO_HOSTNAME" }}
+            {{- $userHostnameOverride = true }}
+          {{- end }}
+        {{- end }}
+        {{- if and .Values.falco.hostnameEnv (not $userHostnameOverride) }}
         - name: FALCO_HOSTNAME
           valueFrom:
             fieldRef:
               fieldPath: spec.nodeName
+        {{- end }}
```
c2ndev, line 102: Same here: the override logic looks good to me, but this should read the chart-level value, not `.Values.falco`.

Proposed final shape:

```gotemplate
{{- if and .Values.falcoHostnameEnv (not $userHostnameOverride) }}
```

Can we also add a small unit test for the default env, `extra.env` override, explicit disable, and to verify this key does not end up in the rendered `falco.yaml`?

charts/falco/README.md
```diff
@@ -661,7 +661,7 @@ The following table lists the main configurable parameters of the falco chart v8
...
...ue":{"capacity":0},"plugins":[],"plugins_hostinfo":true,"priority":"debug","program_output":{"enabled":false,"keep_alive":false,"program":"jq '{text: .output}' | curl -d @- -X POST https://hooks.slack.com/services/XXX"},"rule_matching":"first","rules_files":["/etc/falco/falco_rules.yaml","/etc/falco/falco_rules.local.yaml","/etc/falco/rules.d"],"stdout_output":{"enabled":true},"syscall_event_drops":{"actions":["log","alert"],"max_burst":1,"rate":0.03333,"simulate_drops":false,"threshold":0.1},"syscall_event_timeouts":{"max_consecutives":1000},"syslog_output":{"enabled":true},"time_format_iso_8601":false,"watch_config_files":true,"webserver":{"enabled":true,"k8s_healthz_endpoint":"/healthz","listen_address":"0.0.0.0","listen_port":8765,"prometheus_metrics_enabled":false,"ssl_certificate":"/etc/falco/falco.pem","ssl_enabled":false,"threadiness":0}}` | Directly set values in the falco.yaml configuration file.  Please, note that some values may be overridden by other chart settings, for specific features are managed by the chart itself. This includes: - drivers configuration - collectors configuration (this affects container and k8smeta plugins configurations) - metrics configuration |
```
c2ndev: After moving the value, this generated `falco` object should not include `hostnameEnv`.

charts/falco/README.md
```diff
@@ -692,6 +692,7 @@ The following table lists the main configurable parameters of the falco chart v8
...
 | falco.grpc_output.enabled | bool | `false` | Enable gRPC as an output service. |
+| falco.hostnameEnv | bool | `true` | Inject the chart-default `FALCO_HOSTNAME` env var (sourced from `spec.nodeName` via `valueFrom.fieldRef`) into the Falco container. Set to `false` to suppress entirely. Alternatively, override the value by adding a `FALCO_HOSTNAME` entry to `extra.env` -- the chart auto-detects this and suppresses the default, avoiding a Kubernetes strategic-merge-patch collision on `env[].name` during UPDATE that otherwise produces an invalid EnvVar with both `value` and `valueFrom` set. |
```
c2ndev: Same as above. This should become the chart-level key.

```suggestion
| falcoHostnameEnv | bool | `true` | Inject the chart-default `FALCO_HOSTNAME` env var. If `extra.env` contains `FALCO_HOSTNAME`, the user-provided value wins. |
```

charts/falco/CHANGELOG.md
```diff
@@ -3,6 +3,10 @@
 This file documents all notable changes to Falco Helm Chart. The release
 numbering uses [semantic versioning](http://semver.org).
 
+## v8.0.3
+
+* Make `FALCO_HOSTNAME` env var conditional via new `falco.hostnameEnv` values key (default `true`, backward-compatible). Auto-suppresses the chart default when a user-provided `FALCO_HOSTNAME` is present in `extra.env`, eliminating Kubernetes strategic-merge-patch collisions on UPDATE. Refs falcosecurity/falco#3397.
```
c2ndev: Can we update the key name here too?

```suggestion
* Make chart-injected `FALCO_HOSTNAME` overridable via `extra.env` and suppressible via `falcoHostnameEnv=false`; default behavior is unchanged.
```

charts/falco/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: falco
-version: 8.0.2
+version: 8.0.3
```
c2ndev: `falco-8.0.3` is already published from `master`, so after rebasing this PR should not keep `8.0.3`. Please bump `charts/falco/Chart.yaml` to the next Falco chart patch version, move this changelog entry under that same version, and run `make docs-falco` so `charts/falco/README.md` is regenerated.

c2ndev (commented):
Hey @dunkin0486

The goal makes sense to me. I left a few comments on the values shape before merging.

charts/falco/templates/pod-template.tpl
dunkin0486: Done and tests added

charts/falco/values.yaml
dunkin0486: Done

charts/falco/Chart.yaml
dunkin0486: done

charts/falco/README.md
dunkin0486: done

charts/falco/README.md
dunkin0486: done

charts/falco/CHANGELOG.md
dunkin0486: Done

leogr:
Hey @dunkin0486 

> * `FALCO_K8S_NODE_NAME` immediately below has the same `valueFrom` pattern and would hit the same K8s limitation if anyone tried to override it. Kept out of scope here to keep this PR small; happy to file a follow-up if the maintainers want symmetric treatment.

 Good idea :+1:

leogr (approved):

PR #994, opened 2026-04-20
fix(falco): honor plugins_hostinfo for proc-fs mount

derHahn:
**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

/kind chart-release

/area falco-chart

> /area falcosidekick-chart

> /area falco-talon-chart

> /area event-generator-chart

Fixes #

`falco.plugins_hostinfo` is documented as disabling plugin host-info support and
dropping the `proc-fs` hostPath requirement for plugin-only deployments.
The template currently still emits `proc-fs` unconditionally in the Falco pod
mount/volume path, which causes policy failures in restricted clusters.
**Change**
`proc-fs` is now rendered only when:
- `driver.enabled = true`, or
- `falco.plugins_hostinfo = true`
This preserves syscall-source behavior and makes plugin-only behavior match docs.

**Checklist**

- [x] Chart Version bumped
- [x] CHANGELOG.md updated


charts/falco/templates/pod-template.tpl
```diff
@@ -1,4 +1,5 @@
 {{- define "falco.podTemplate" -}}
+{{- $procMountEnabled := or .Values.driver.enabled .Values.falco.plugins_hostinfo -}}
```
c2ndev: This condition is correct, but I would not keep it as a local variable in `pod-template.tpl`. Please delete this line after moving the predicate to a helper.

Please move the predicate to `_helpers.tpl`, using the same string-returning helper style already used by `driverLoader.enabled` and `falco.sysfsMount.enabled`:

```gotemplate
{{/*
Return "true" if proc-fs should be mounted in the Falco pod.
*/}}
{{- define "falco.procfsMount.enabled" -}}
{{- if or .Values.driver.enabled .Values.falco.plugins_hostinfo -}}
true
{{- else -}}
false
{{- end -}}
{{- end -}}
```

Then use that helper in both render points below and add unit coverage for the matrix:

- default / `driver.enabled=true`: `proc-fs` mount and volume are still present
- `driver.enabled=false`, `falco.plugins_hostinfo=true`: `proc-fs` is still present
- `driver.enabled=false`, `falco.plugins_hostinfo=false`: `proc-fs` is absent from both the Falco container mounts and pod volumes

charts/falco/templates/pod-template.tpl
```diff
@@ -151,10 +152,14 @@ spec:
       {{- end }}
         - mountPath: /root/.falco
           name: root-falco-fs
+        {{- if $procMountEnabled }}
```
c2ndev: Use the helper here instead of the local variable.

```suggestion
        {{- if eq (include "falco.procfsMount.enabled" .) "true" }}
```

charts/falco/templates/pod-template.tpl
```diff
@@ -272,9 +277,11 @@ spec:
       hostPath:
         path: {{ .Values.driver.sysfsMountPath }}
     {{- end }}
+    {{- if $procMountEnabled }}
```
c2ndev: Same here, so the mount and volume are guarded by the same chart helper.

```suggestion
    {{- if eq (include "falco.procfsMount.enabled" .) "true" }}
```

charts/falco/README.md
```diff
@@ -723,7 +723,7 @@ The following table lists the main configurable parameters of the falco chart v8
...
-| falco.plugins_hostinfo | bool | `true` | Control host info support for plugins. When `false`, it disables host info support for source plugins that DO NOT generate raw events from the libscap event table or for plugins that DO NOT parse raw events generated by drivers, effectively dropping the `proc-fs` hostPath volume requirement for them: https://github.com/falcosecurity/charts/blob/bd57711e7c8e00919ea288716e0d9d5fdad8867e/charts/falco/templates/pod-template.tpl#L302-L304 |
+| falco.plugins_hostinfo | bool | `true` | Control host info support for plugins. `proc-fs` is rendered when `driver.enabled=true` or `falco.plugins_hostinfo=true`; therefore, setting `driver.enabled=false` and `falco.plugins_hostinfo=false` drops the `proc-fs` hostPath requirement for plugin-only deployments. |
```
c2ndev: This is the right wording, but the source comment in `values.yaml` still has the old text and stale permalink. Since this table is generated, I would update the `falco.plugins_hostinfo` comment in `values.yaml` and regenerate the README instead of editing only this row.

Something like:

```yaml
# -- Control host info support for plugins. `proc-fs` is rendered when
# `driver.enabled=true` or `falco.plugins_hostinfo=true`; therefore, setting
# both `driver.enabled=false` and `falco.plugins_hostinfo=false` drops the
# `proc-fs` hostPath requirement for plugin-only deployments.
```

This will also refresh the README chart version line from the final `Chart.yaml` version.

Useful command for this part:

```bash
make docs-falco
```

charts/falco/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: falco
-version: 8.0.2
+version: 8.0.3
```
c2ndev: Since this changes the chart version, please keep `/kind chart-release` or add the label. `falco-8.0.3` is already on `master` and published, and this PR is currently conflicting with `master`; after rebasing, please bump to the next Falco chart patch version and put this changelog entry under that final version.

c2ndev (commented):
Hey @derHahn, thanks for fixing this.

The direction looks good to me and keeps the current driver behavior. I left a few comments to make it release-ready.

charts/falco/templates/pod-template.tpl
derHahn: Thanks for your suggestions @c2ndev 
Fixed as requested. Created a new test file for this.

c2ndev:
Thanks @derHahn for the update, the implementation looks good to me now. The helper approach is aligned with the chart style, and the rendered behavior is correct.

Before merging, could you please just align the release metadata as well? The changelog entry is under `v8.0.4`, but `Chart.yaml` is still `8.0.3`, so we should sync `Chart.yaml`, `CHANGELOG.md`, and add the `kind/chart-release` label with the final version for this PR.

One last small test improvement: can you add the `driver.enabled=true` + `falco.plugins_hostinfo=false` case too? The current tests cover the plugin-only behavior; this extra case explicitly protects the driver side of the helper predicate.

After that, I think this is good from my side.

derHahn:
> Thanks @derHahn for the update, the implementation looks good to me now. The helper approach is aligned with the chart style, and the rendered behavior is correct.
> 
> Before merging, could you please just align the release metadata as well? The changelog entry is under `v8.0.4`, but `Chart.yaml` is still `8.0.3`, so we should sync `Chart.yaml`, `CHANGELOG.md`, and add the `kind/chart-release` label with the final version for this PR.
> 
> One last small test improvement: can you add the `driver.enabled=true` + `falco.plugins_hostinfo=false` case too? The current tests cover the plugin-only behavior; this extra case explicitly protects the driver side of the helper predicate.
> 
> After that, I think this is good from my side.

Thanks again, @c2ndev! And sorry for missing these earlier! I've added the label now as well.

ekoops (changes requested):
Hey! Changes look good. Could you please squash all the commits and use conventional commit syntax con both PR and commit titles?

derHahn:
> Hey! Changes look good. Could you please squash all the commits and use conventional commit syntax con both PR and commit titles?

Thank you @ekoops, all done! May I ask why you don't use squash & merge?

ekoops:
> > Hey! Changes look good. Could you please squash all the commits and use conventional commit syntax con both PR and commit titles?
> 
> Thank you @ekoops, all done! May I ask why you don't use squash & merge?

Good question. We discussed morning about it, but changing the merge strategy would mean changing it on each and every falcosecurity repo. Since we are approaching Falco release, it is better to address this later, but yes, it would make sense :smile:

leogr:
Hey @derHahn 

Can you rebase this? Then I'm ready to approve. Thanks!

charts/falco/CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ numbering uses [semantic versioning](http://semver.org).
 
 ## v8.0.4
 
+* Make chart-injected `FALCO_HOSTNAME` overridable via `extra.env` and suppressible via `falcoHostnameEnv=false`; default behavior is unchanged.
+* Honor `falco.plugins_hostinfo` for `proc-fs` mount and volume generation in plugin-only deployments, while keeping existing behavior when `driver.enabled=true`
 * Make chart-injected `FALCO_HOSTNAME` overridable via `extra.env` and suppressible via `falcoHostnameEnv=false`; default behavior is unchanged.
```
c2ndev: Sorry for stressing you, I think during the rebase something went duplicated. Can you make this final check please? 🙏

derHahn: Oh my, sorry for the oversight and thanks for pointing it out! Should be fine now!

ekoops:
Hey, sorry to bother you, but given the change, the chart version need to be bumped :pensive: : https://github.com/falcosecurity/charts/actions/runs/25783514385/job/75739501272?pr=994

ekoops (approved):
Thank you! Changes look good 😄 
/approve

PR #1029, opened 2026-06-15
fix(falco-talon): fix missing namespace in the Secret metadata

### LINKED ISSUE #1028: falco-talon: config Secret is missing namespace in its metadata
**Describe the bug**

In the falco-talon chart, the config Secret template (`charts/falco-talon/templates/secrets.yaml`) does not set `namespace` in its metadata.
The other manifests in the falco-talon chart all set it explicitly.
The config Secret seems to be the only one missing it, which looks like it may have been left out unintentionally.

When the rendered manifests are applied directly (e.g. GitOps/ArgoCD or helm template | kubectl apply), the Secret ends up without a namespace, unlike the other resources in this chart. (Plain helm install is unaffected.)
https://github.com/falcosecurity/charts/blob/53586de4fb9d8d02006131ade702b161cd7e06e3/charts/falco-talon/templates/secrets.yaml#L3-L6

**How to reproduce it**

Render the chart into a namespace and inspect the Secret with yq:
```
$ helm repo add falcosecurity https://falcosecurity.github.io/charts
$ helm repo update falcosecurity
$ helm template rel falcosecurity/falco-talon --version 0.4.0 -n falco \
      | yq 'select(.kind == "Secret") | {"kind": .kind, "name": .metadata.name, "namespace": .metadata.namespace}'
kind: Secret
name: falco-talon-config
namespace: null
```

**Expected behaviour**

The config Secret should set `namespace: {{ .Release.Namespace }}` in its metadata.

zeroalphat:
**What type of PR is this?**

/kind bug
/kind chart-release

/area falco-talon-chart

The Secret does not set a namespace. As a result, when the manifests are rendered and then applied, the Secret is created without a namespace and the Pod cannot reference it, so the Pod fails to start.
This PR sets the namespace on the Secret, which resolves the issue.
For more details, refer to #1028.

Fixes #1028

**Checklist**

- [x] Chart Version bumped
- [x] CHANGELOG.md updated

### DIFF
--- a/charts/falco-talon/CHANGELOG.md
+++ b/charts/falco-talon/CHANGELOG.md
@@ -3,6 +3,10 @@
 This file documents all notable changes to Falco Talon Helm Chart. The release
 numbering uses [semantic versioning](http://semver.org).
 
+## 0.4.1 - 2026-06-15
+
+- fix missing namespace in the Secret metadata
+
 ## 0.4.0 - 2025-05-02
 
 - Allow specifying folder annotation for grafana dashboards
--- a/charts/falco-talon/Chart.yaml
+++ b/charts/falco-talon/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: 0.3.0
 description: React to the events from Falco
 name: falco-talon
-version: 0.4.0
+version: 0.4.1
 keywords:
   - falco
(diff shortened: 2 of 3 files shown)


ekoops (approved):
Thank you for this contribution 🙏 
/approve

PR #1037, opened 2026-07-02
feat: introducing default networkPolicy using "defaultNetworkPolicy" dedicated values section

heitzflorian:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

/kind chart-release

> /area falco-chart

/area falcosidekick-chart

> /area falco-talon-chart

> /area event-generator-chart

> /area falco-operator-chart

> /area automation

By default no network policy is set by falcosidekick helm chart.
As falco is ran by default as a DaemonSet and requires the securityContext: privileged: true (equivalent to docker --privileged ), it can become a suitable and preferred target for an attacker to get in.

This PR introduce a "default" networkPolicy that can deployed and activated during helm install by setting `defaultNetworkPolicy.enabled: true`.

Extra parameters `defaultNetworkPolicy.webuiIngressNetworkPoliciesNamespaceSelector` and `defaultNetworkPolicy.webuiIngressNetworkPoliciesPodSelector` are configurable to tailor the selectors used to apply an additional and dedicated networkPolicy targeting ingress-controller pods (conditioned by its activation, see `webui.enabled=true` and `webui.ingress.enabled=true`).

Fixes #979 

Non regression tests Matrix done, and results:
| defaultNetworkPolicy.enabled | webui.enabled  | webui.ingress.enabled | networkPolicy rendered | sidekick-core                                                           | sidekick-ui                                                   | sidekick-ui-redis                                   |
| ---------------------------- | -------------- | --------------------- | ---------------------- | ----------------------------------------------------------------------- | ------------------------------------------------------------- | --------------------------------------------------- |
| false                        | (true / false) | (true / false)        | NO                     | none                                                                    | none                                                          | none                                                |
| true                         | false          | (true / false)        | YES                    | <pre>ingress TCP 2801, TCP 2810</pre><pre>egress UDP 53</pre>           | none                                                          | none                                                |
| true                         | true           | false                 | YES                    | <pre>ingress TCP 2801, TCP 2810</pre><pre>egress UDP 53, TCP 2802</pre> | <pre>ingress TCP 2802*</pre><pre>egress UDP 53, TCP 6379</pre> | <pre>ingress TCP 6379</pre><pre>egress UDP 53</pre> |
| true                         | true           | true                  | YES                    | <pre>ingress TCP 2801, TCP 2810</pre><pre>egress UDP 53, TCP 2802</pre> | <pre>ingress TCP 2802**</pre><pre>egress UDP 53, TCP 6379</pre> | <pre>ingress TCP 6379</pre><pre>egress UDP 53</pre> |

*tcp Port 2802 open only for  sidekick-core
**tcp Port 2802 open for both sidekick-core and ingress

Detailed results in GIST link: https://gist.github.com/heitzflorian/de36a2756471e52bf49569c6f1197241

**Checklist**

- [x] Chart Version bumped
- [x] CHANGELOG.md updated

### DIFF
--- a/charts/falcosidekick/templates/network-policy-core.yaml
+++ b/charts/falcosidekick/templates/network-policy-core.yaml
@@ -1,0 +1,59 @@
+{{- if .Values.defaultNetworkPolicy.enabled -}}
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: {{ include "falcosidekick.fullname" . }}-core-network-policy
+  namespace: {{ .Release.Namespace }}
+spec:
+  podSelector:
+    matchLabels:
+      app.kubernetes.io/component: core
+      app.kubernetes.io/instance: falco
+      app.kubernetes.io/name: falcosidekick
+  policyTypes:
+    - Ingress
+    - Egress
+  ingress:
+    # Accept Traffic on ports 2801 and 2810 from falco
+    - from:
+        - namespaceSelector:
+            matchLabels:
+              kubernetes.io/metadata.name: falco
+          podSelector:
+            matchLabels:
+              app.kubernetes.io/instance: falco
+              app.kubernetes.io/name: falco
+          podSelector:
+      ports:
+        # http-notls
+        - port: 2801
+          protocol: TCP
+        # http-notls
+        - port: 2810
+          protocol: TCP
+  egress:
+    # Allow kube-dns resolution
+    - to:
+        - namespaceSelector:
+            matchLabels:
+              kubernetes.io/metadata.name: kube-system
+          podSelector:
+            matchLabels:
+              k8s-app: kube-dns
+      ports:
+        - port: 53
+          protocol: UDP
+    {{- if .Values.webui.enabled }}
+    # Allow sending POST to sidekick-ui to ingest data
+    - to:
+        - podSelector:
+            matchLabels:
+             app.kubernetes.io/component: ui
+             app.kubernetes.io/instance: falco
+             app.kubernetes.io/name: falcosidekick
+      ports:
+        # sidekick-ui
+        - port: 2802
+          protocol: TCP
+    {{- end -}}
+{{- end -}}
--- a/charts/falcosidekick/templates/network-policy-redis.yaml
+++ b/charts/falcosidekick/templates/network-policy-redis.yaml
@@ -1,0 +1,45 @@
+{{- if .Values.defaultNetworkPolicy.enabled -}}
+{{- if .Values.webui.enabled -}}
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: {{ include "falcosidekick.fullname" . }}-ui-redis-network-policy
+  namespace: {{ .Release.Namespace }}
+spec:
+  podSelector:
+    matchLabels:
+      app.kubernetes.io/component: ui-redis
+      app.kubernetes.io/instance: falco
+      app.kubernetes.io/name: falcosidekick
+  policyTypes:
+    - Ingress
+    - Egress
+  ingress:
+    # Accept Traffic on REDIS port 6379 from falcosidekick-ui
+    - from:
+        - namespaceSelector:
+            matchLabels:
+              kubernetes.io/metadata.name: falco
+          podSelector:
+            matchLabels:
+              app.kubernetes.io/component: ui
+              app.kubernetes.io/instance: falco
+              app.kubernetes.io/name: falcosidekick
+          podSelector:
+      ports:
+        - port: 6379
+          protocol: TCP
+  egress:
+    # Allow kube-dns resolution
+    - to:
+        - namespaceSelector:
+            matchLabels:
+              kubernetes.io/metadata.name: kube-system
+          podSelector:
+            matchLabels:
+              k8s-app: kube-dns
+      ports:
+        - port: 53
+          protocol: UDP
+{{- end -}}
+{{- end -}}
--- a/charts/falcosidekick/templates/network-policy-ui.yaml
+++ b/charts/falcosidekick/templates/network-policy-ui.yaml
@@ -1,0 +1,101 @@
+{{- if .Values.defaultNetworkPolicy.enabled -}}
+{{- if .Values.webui.enabled -}}
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: {{ include "falcosidekick.fullname" . }}-ui-network-policy
+  namespace: {{ .Release.Namespace }}
+spec:
+  podSelector:
+    matchLabels:
+      app.kubernetes.io/component: ui
+      app.kubernetes.io/instance: falco
+      app.kubernetes.io/name: falcosidekick
+  policyTypes:
+    - Ingress
+    - Egress
+  ingress:
+    # Accept Traffic on port 2802 from falcosidekick-core
+    - from:
+        - podSelector:
+            matchLabels:
+              app.kubernetes.io/component: core
+              app.kubernetes.io/instance: falco
+              app.kubernetes.io/name: falcosidekick
+      ports:
+        - port: 2802
+          protocol: TCP
+  egress:
+    # Allow Traffic on REDIS port 6379
+    - to:
+      - podSelector:
+          matchLabels:
+            app.kubernetes.io/component: ui-redis
+            app.kubernetes.io/instance: falco
+            app.kubernetes.io/name: falcosidekick
+      ports:
+        - port: 6379
+          protocol: TCP
+    # Allow kube-dns resolution
+    - to:
+        - namespaceSelector:
+            matchLabels:
+              kubernetes.io/metadata.name: kube-system
+          podSelector:
+            matchLabels:
+              k8s-app: kube-dns
+      ports:
+        - port: 53
+          protocol: UDP
+{{- if .Values.webui.ingress.enabled }}
+---
+{{- if or (.Values.defaultNetworkPolicy.webuiIngressNetworkPolicyNamespaceSelector) (.Values.defaultNetworkPolicy.webuiIngressNetworkPolicyPodSelector) }}
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: {{ include "falcosidekick.fullname" . }}-ui-network-policy-webui-ingress
+  namespace: {{ .Release.Namespace }}
+spec:
+  podSelector:
+    matchLabels:
+      app.kubernetes.io/component: ui
+      app.kubernetes.io/instance: falco
+      app.kubernetes.io/name: falcosidekick
+  policyTypes:
+    - Ingress
+  ingress:
+    # Accept Traffic on port 2802 from ingress controller
+    - from:
+        {{- if and (.Values.defaultNetworkPolicy.webuiIngressNetworkPolicyNamespaceSelector) (.Values.defaultNetworkPolicy.webuiIngressNetworkPolicyPodSelector) }}
+        - namespaceSelector:
+            matchLabels:
+              {{- with .Values.defaultNetworkPolicy.webuiIngressNetworkPolicyNamespaceSelector }}
+              {{- toYaml . | nindent 14 }}
+              {{- end }}
+          podSelector:
+            matchLabels:
+              {{- with .Values.defaultNetworkPolicy.webuiIngressNetworkPolicyPodSelector }}
+              {{- toYaml . | nindent 14 }}
+              {{- end }}
+        {{- end }}
+        {{- if and (.Values.defaultNetworkPolicy.webuiIngressNetworkPolicyNamespaceSelector) (not .Values.defaultNetworkPolicy.webuiIngressNetworkPolicyPodSelector) }}
+        - namespaceSelector:
+            matchLabels:
+              {{- with .Values.defaultNetworkPolicy.webuiIngressNetworkPolicyNamespaceSelector }}
+              {{- toYaml . | nindent 14 }}
+              {{- end }}
+        {{- end }}
+        {{- if and (.Values.defaultNetworkPolicy.webuiIngressNetworkPolicyPodSelector) (not .Values.defaultNetworkPolicy.webuiIngressNetworkPolicyNamespaceSelector) }}
+        - podSelector:
+            matchLabels:
+              {{- with .Values.defaultNetworkPolicy.webuiIngressNetworkPolicyPodSelector }}
+              {{- toYaml . | nindent 14 }}
+              {{- end }}
+        {{- end }}
+      ports:
+        - port: 2802
+          protocol: TCP
+{{- end -}}
+{{- end -}}
+{{- end -}}
+{{- end -}}
--- a/charts/falcosidekick/values.yaml
+++ b/charts/falcosidekick/values.yaml
@@ -1172,6 +1172,20 @@ ingress:
   #    hosts:
   #      - chart-example.local
 
+# -- Default networkPolicy that comes out of the box with falcosidekick if you don't have any specific needs.
+defaultNetworkPolicy:
+  enabled: false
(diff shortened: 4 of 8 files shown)


charts/falcosidekick/templates/network-policy-core.yaml
```diff
@@ -0,0 +1,59 @@
...
+kind: NetworkPolicy
+metadata:
+  name: {{ include "falcosidekick.fullname" . }}-core-network-policy
+  namespace: {{ .Release.Namespace }}
+spec:
+  podSelector:
+    matchLabels:
+      app.kubernetes.io/component: core
+      app.kubernetes.io/instance: falco
+      app.kubernetes.io/name: falcosidekick
```
c2ndev, line 12:
```suggestion
      app.kubernetes.io/component: core
      {{- include "falcosidekick.selectorLabels" . | nindent 6 }}
```

`app.kubernetes.io/instance` is set to `{{ .Release.Name }}` by the chart helpers, so hardcoding `falco` makes the policy select no pods unless the release happens to be named `falco`. I rendered with `helm template myrelease .` and the selector still says `falco`, so the policy silently does nothing. Same for `app.kubernetes.io/name`, which can change via `nameOverride`.

Could we use the `falcosidekick.selectorLabels` helper plus the component label, like the Deployments and Services do? The same applies to every pod selector block in the three new templates (ui, ui-redis, and the ingress/egress peers).

charts/falcosidekick/templates/network-policy-core.yaml
```diff
@@ -0,0 +1,59 @@
...
+    # Accept Traffic on ports 2801 and 2810 from falco
+    - from:
+        - namespaceSelector:
+            matchLabels:
+              kubernetes.io/metadata.name: falco
+          podSelector:
+            matchLabels:
+              app.kubernetes.io/instance: falco
+              app.kubernetes.io/name: falco
+          podSelector:
```
c2ndev, line 26:
```suggestion
```

This stray `podSelector:` is a duplicate key in the same mapping. The YAML-to-JSON conversion used by kubectl and Helm takes the last key, so it overrides the real selector above with `podSelector: null`. I verified with `kubectl create --dry-run=client -o yaml`: the applied rule becomes "any pod in the falco namespace", silently dropping the Falco pod restriction. Same stray key in `network-policy-redis.yaml` line 28.

charts/falcosidekick/templates/network-policy-core.yaml
```diff
@@ -0,0 +1,59 @@
...
+  ingress:
+    # Accept Traffic on ports 2801 and 2810 from falco
+    - from:
+        - namespaceSelector:
+            matchLabels:
+              kubernetes.io/metadata.name: falco
+          podSelector:
+            matchLabels:
+              app.kubernetes.io/instance: falco
+              app.kubernetes.io/name: falco
```
c2ndev, line 25: The source selectors for Falco are hardcoded here: namespace `falco` and `app.kubernetes.io/instance: falco`. Falco's instance label is its release name, and nothing guarantees the namespace is called `falco` (the docs just use it in examples). With any other layout, Falco's POSTs to 2801 get dropped as soon as this policy is enabled, and it will look like falcosidekick is broken.

What about making these configurable, e.g. `defaultNetworkPolicy.falcoNamespaceSelector` / `falcoPodSelector`, defaulting to a same-namespace podSelector? That would cover the common case where falcosidekick is installed as a dependency of the falco chart and shares its namespace and release name.

charts/falcosidekick/templates/network-policy-core.yaml
```diff
@@ -0,0 +1,59 @@
...
+          podSelector:
+      ports:
+        # http-notls
+        - port: 2801
+          protocol: TCP
+        # http-notls
+        - port: 2810
+          protocol: TCP
+  egress:
```
c2ndev, line 34: This is the one I would like to discuss before merging. With `policyTypes: [Ingress, Egress]`, everything not listed is denied, and the only egress allowed is DNS and the UI. Falcosidekick's whole job is pushing events to external outputs (Slack, Elasticsearch, webhooks, and so on), so enabling `defaultNetworkPolicy.enabled=true` breaks every configured output.

Could we either restrict the default policy to Ingress only, or add a configurable egress allowlist (plus an `egress.enabled` toggle)? Also a side question: should DNS allow TCP 53 too? Truncated responses fall back to TCP in some clusters.

charts/falcosidekick/templates/network-policy-core.yaml
```diff
@@ -0,0 +1,59 @@
...
+              app.kubernetes.io/instance: falco
+              app.kubernetes.io/name: falco
+          podSelector:
+      ports:
+        # http-notls
+        - port: 2801
+          protocol: TCP
+        # http-notls
+        - port: 2810
+          protocol: TCP
```
c2ndev, line 33: Small things on the ports: the comment on 2801 says `http-notls`, but 2801 is the main HTTP port; 2810 is the no-TLS one. Also 2810 only exists on the pod when `config.tlsserver.deploy=true`, and its value comes from `config.tlsserver.notlsport`, so it may be worth templating these from the same values the Deployment and Service use, to avoid drift when users override them.

charts/falcosidekick/templates/network-policy-core.yaml
```diff
@@ -0,0 +1,59 @@
+{{- if .Values.defaultNetworkPolicy.enabled -}}
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: {{ include "falcosidekick.fullname" . }}-core-network-policy
+  namespace: {{ .Release.Namespace }}
```
c2ndev, line 6:
```suggestion
metadata:
  name: {{ include "falcosidekick.fullname" . }}-core-network-policy
  namespace: {{ .Release.Namespace }}
  labels:
    {{- include "falcosidekick.labels" . | nindent 4 }}
    app.kubernetes.io/component: core
    {{- with .Values.customLabels }}
    {{- toYaml . | nindent 4 }}
    {{- end }}
  annotations:
    {{- with .Values.customAnnotations }}
    {{- toYaml . | nindent 4 }}
    {{- end }}
```

All other resources in this chart attach the common labels, the component label, and `customLabels`/`customAnnotations`. Could we do the same for the three new NetworkPolicies?

charts/falcosidekick/templates/network-policy-redis.yaml
```diff
@@ -0,0 +1,45 @@
+{{- if .Values.defaultNetworkPolicy.enabled -}}
+{{- if .Values.webui.enabled -}}
```
c2ndev, line 2:
```suggestion
{{- if .Values.webui.redis.enabled -}}
```

The Redis Service and workload only render when `webui.redis.enabled` is true (Redis can be disabled in favor of `webui.externalRedis`). Gating on `webui.enabled` alone renders a policy that selects nothing in that case.

charts/falcosidekick/templates/network-policy-redis.yaml
```diff
@@ -0,0 +1,45 @@
...
+    - from:
+        - namespaceSelector:
+            matchLabels:
+              kubernetes.io/metadata.name: falco
+          podSelector:
+            matchLabels:
+              app.kubernetes.io/component: ui
+              app.kubernetes.io/instance: falco
+              app.kubernetes.io/name: falcosidekick
+          podSelector:
```
c2ndev, line 28:
```suggestion
        - podSelector:
            matchLabels:
              app.kubernetes.io/component: ui
              {{- include "falcosidekick.selectorLabels" . | nindent 14 }}
```

The UI always runs in the release namespace, so the namespaceSelector on a hardcoded `falco` namespace is not needed and breaks other layouts. A plain podSelector with the chart helpers covers it, and this also removes the stray duplicate `podSelector:` key at line 28.

charts/falcosidekick/templates/network-policy-ui.yaml
```diff
@@ -0,0 +1,101 @@
...
+    # Allow Traffic on REDIS port 6379
+    - to:
+      - podSelector:
+          matchLabels:
+            app.kubernetes.io/component: ui-redis
+            app.kubernetes.io/instance: falco
+            app.kubernetes.io/name: falcosidekick
+      ports:
+        - port: 6379
+          protocol: TCP
```
c2ndev, line 38: When `webui.externalRedis.enabled=true` the UI connects to an external endpoint on `webui.externalRedis.port`, and this rule only allows in-cluster traffic to the ui-redis pods on a hardcoded 6379, so the UI cannot reach its database once the policy is on. The in-cluster port should probably come from `webui.redis.service.port` too.

Could we handle the external Redis case, or at least document that the default policy does not support it?

charts/falcosidekick/templates/network-policy-ui.yaml
```diff
@@ -0,0 +1,101 @@
...
+              kubernetes.io/metadata.name: kube-system
+          podSelector:
+            matchLabels:
+              k8s-app: kube-dns
+      ports:
+        - port: 53
+          protocol: UDP
+{{- if .Values.webui.ingress.enabled }}
+---
+{{- if or (.Values.defaultNetworkPolicy.webuiIngressNetworkPolicyNamespaceSelector) (.Values.defaultNetworkPolicy.webuiIngressNetworkPolicyPodSelector) }}
```
c2ndev, line 52: Two things here. The `---` sits outside the inner `if`, so an empty YAML document is emitted when both selectors are unset (cosmetic, moving it inside the `if or` fixes it).

More importantly: with the default empty selectors and `webui.ingress.enabled=true`, no allow rule is rendered for the ingress controller, so the base UI policy blocks it and the ingress silently stops working. The same applies when the UI is exposed via a LoadBalancer or NodePort Service without an ingress. Could we fail the render with `required`, or at least call this out loudly in values.yaml and the README?

charts/falcosidekick/values.yaml
```diff
@@ -1172,6 +1172,20 @@ ingress:
   #    hosts:
   #      - chart-example.local
 
+# -- Default networkPolicy that comes out of the box with falcosidekick if you don't have any specific needs.
+defaultNetworkPolicy:
+  enabled: false
+  # The following lines will work only if webui ingress is enabled (webui.ingress.enabled = true)
+  webuiIngressNetworkPolicyNamespaceSelector: {}
+  webuiIngressNetworkPolicyPodSelector: {}
+  # Full Example with nginx (deprecrated) below.
```
c2ndev, line 1181: Since these are the first NetworkPolicies in our charts, this naming sets a precedent. Most charts in the wider ecosystem seem to use a `networkPolicy` section with `networkPolicy.enabled`; should we align on that instead of `defaultNetworkPolicy`? The nested names could also get shorter, e.g. `networkPolicy.ingressController.namespaceSelector` / `podSelector`.

Also a small typo below: "deprecrated" should be "deprecated".

c2ndev (commented):
Thanks for this contribution! The non-regression test matrix in the description is much appreciated.

I left inline some comments.

charts/falcosidekick/templates/network-policy-core.yaml
heitzflorian: I accepted this suggestion

charts/falcosidekick/templates/network-policy-core.yaml
heitzflorian: I accepted this suggestion

charts/falcosidekick/templates/network-policy-redis.yaml
heitzflorian: I accepted this suggestion

charts/falcosidekick/templates/network-policy-redis.yaml
heitzflorian: I accepted this suggestion

charts/falcosidekick/templates/network-policy-core.yaml
heitzflorian: I accepted this suggestion

charts/falcosidekick/templates/network-policy-core.yaml
heitzflorian: Hi @c2ndev,

sorry for th "http-notls" comment which result from a dirty copy-paste without any edits.
Regarding the suggested behavior, it can be something like this:

```yaml
      ports:
        # http
        - port: 2801
          protocol: TCP
        {{- if and (not .Values.config.tlsserver.deploy) (ne .Values.config.tlsserver.notlsport "") }}
        # http-notls
        - port: {{ .Values.config.tlsserver.notlsport }}
          protocol: TCP
        {{- end }}
```

See following GIST for test logs: https://gist.github.com/heitzflorian/1c7285fb11713ddc6c38dcad4dd122cd .

charts/falcosidekick/templates/network-policy-ui.yaml
heitzflorian: Hi @c2ndev,

Regarding the suggested behavior, it can be something like this:
```yaml
  egress:
    {{- if or .Values.webui.redis.enabled (.Values.webui.externalRedis.enabled) }}
    # Allow Traffic on REDIS port
    - to:
      - podSelector:
          matchLabels:
            app.kubernetes.io/component: ui-redis
            app.kubernetes.io/instance: falco
            app.kubernetes.io/name: falcosidekick
      ports:
        - protocol: TCP
        {{- if .Values.webui.redis.enabled }}
          port: {{ .Values.webui.redis.service.port }}
        {{- else if .Values.webui.externalRedis.enabled }}
          port: {{ .Values.webui.externalRedis.port }}
        {{- end }}
    {{- end }}
```
See following GIST files for test logs: 
* [file-network-policy-ui-yaml](https://gist.github.com/heitzflorian/1c7285fb11713ddc6c38dcad4dd122cd#file-network-policy-ui-yaml)
* [helm_template_netpol_ui_testing.log](https://gist.github.com/heitzflorian/1c7285fb11713ddc6c38dcad4dd122cd#file-helm_template_netpol_ui_testing-log)

charts/falcosidekick/templates/network-policy-core.yaml
heitzflorian: Regarding the suggested behavior, it can be something like this:

`values.yml`:
```yaml
# -- Default networkPolicy that comes out of the box with falcosidekick if you don't have any specific needs.
defaultNetworkPolicy:
  enabled: false
  falcoNamespaceSelector:
    kubernetes.io/metadata.name: falco
  falcoPodSelector:
    app.kubernetes.io/instance: falco
    app.kubernetes.io/name: falco
  # The following lines will work only if webui ingress is enabled (webui.ingress.enabled = true)
  webuiIngressNetworkPolicyNamespaceSelector: {}
  webuiIngressNetworkPolicyPodSelector: {}
  # Full Example with nginx (deprecrated) below.
  # webuiIngressNetworkPolicyNamespaceSelector:
  #  kubernetes.io/metadata.name: ingress-nginx
  # webuiIngressNetworkPolicyPodSelector:
  #  app.kubernetes.io/component: controller
  #  app.kubernetes.io/instance: ingress-nginx
  #  app.kubernetes.io/name: ingress-nginx
```

`network-policy-core.yaml`:
```yaml
  ingress:
    # Accept Traffic on http ports from falco
    - from:
        - namespaceSelector:
            matchLabels:
              {{- with .Values.defaultNetworkPolicy.falcoNamespaceSelector }}
              {{- toYaml . | nindent 14 }}
              {{- end }}
          podSelector:
            matchLabels:
              {{- with .Values.defaultNetworkPolicy.falcoPodSelector }}
              {{- toYaml . | nindent 14 }}
              {{- end }}
```
See following GIST for test logs:
* [file-network-policy-core-yaml](https://gist.github.com/heitzflorian/1c7285fb11713ddc6c38dcad4dd122cd#file-network-policy-core-yaml)
* [file-helm_template_netpol_core_selectors_testing-log](https://gist.github.com/heitzflorian/1c7285fb11713ddc6c38dcad4dd122cd#file-helm_template_netpol_core_selectors_testing-log)

PR #1054, opened 2026-09-02
feat(release): support pre-release chart versions

alacuku:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /kind chart-release

> /area falco-chart

> /area falcosidekick-chart

> /area falco-talon-chart

> /area event-generator-chart

> /area falco-operator-chart

> /area automation

Publish charts with a SemVer pre-release suffix (e.g. 1.2.3-rc.1) as proper GitHub pre-releases instead of "Latest release", decided per chart from its packaged metadata rather than a per-push guess.

Fixes #

**Checklist**

- [ ] Chart Version bumped
- [ ] CHANGELOG.md updated

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -45,6 +45,10 @@ jobs:
         with:
           charts_dir: charts
           config: cr.yaml
+          # "Latest" and pre-release status are decided per chart in the
+          # "Publish and Sign OCI Charts" step below, from each package's own
+          # metadata rather than a guess about what this push touched.
+          mark_as_latest: false
         env:
           CR_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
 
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -57,12 +61,53 @@ jobs:
 
       - name: Publish and Sign OCI Charts
         run: |
-          for chart in `find .cr-release-packages -name '*.tgz' -print`; do
-            helm push ${chart} oci://ghcr.io/${GITHUB_REPOSITORY} |& tee helm-push-output.log
-            file_name=${chart##*/}
-            chart_name=${file_name%-*}
+          overall_failed=false
```
leogr, line 65:
```suggestion
          set -o pipefail
          overall_failed=false
```

Since this step does not set `shell: bash`, GitHub runs it with `bash -e {0}` only (no `pipefail`), so the exit status of `helm push ... |& tee helm-push-output.log` below is the one of `tee`. As a result, the `if ! helm push ...` check never fires when `helm push` fails, and the failure would only surface later as a confusing `cosign sign` error with an empty digest.

Alternatively, we can add `shell: bash` to the step, since GitHub then runs it with `-eo pipefail` 👉 https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#exit-codes-and-error-action-preference

.github/workflows/release.yml
```diff
@@ -57,12 +61,53 @@ jobs:
...
+          for chart in .cr-release-packages/*.tgz; do
+            [[ -e "${chart}" ]] || continue
+
+            # Read name/version from the package's own Chart.yaml via `yq`
+            if ! chart_meta=$(helm show chart "${chart}"); then
+              echo "::error::helm show chart failed for ${chart}, skipping"
+              overall_failed=true
+              continue
+            fi
+            read -r chart_name chart_version <<< "$(yq -r '.name + " " + .version' <<< "${chart_meta}")"
```
leogr, line 78:
```suggestion
            read -r chart_name chart_version <<< "$(yq -r '.name + " " + .version' <<< "${chart_meta}")"
            if [[ -z "${chart_name}" || -z "${chart_version}" ]]; then
              echo "::error::could not read chart name/version from ${chart}, skipping"
              overall_failed=true
              continue
            fi
```

Nit. If `yq` fails for any reason, `read` still returns 0 (the here-string always provides a newline), so we would proceed with an empty `chart_name`/`chart_version` and get less obvious errors from `cosign sign` and `gh release edit`. Not a blocker, just for consistency with the `helm show chart` check above.

leogr (approved):

PR #633, opened 2024-02-22
use directory-mapping instead of simple containerd.socket-file-mapping

devfaz:
**What type of PR is this?**

/kind bug

/kind chart-release

**Any specific area of the project related to this PR?**

/area falco-chart

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:

Fixes #632 

**Special notes for your reviewer**:

to allow falco to reconnect if containerd got restarted on host

**Checklist**

- [x] Chart Version bumped
- [x] Variables are documented in the README.md
- [x] CHANGELOG.md updated

Issue #632, opened 2024-02-22
falco will lose connection to containerd.socket if containerd on host gets restarted


devfaz:
**Describe the bug**

If the containerd-process on the host-system gets restarted. Falco will loose the connection to the socket.

**How to reproduce it**

* Increase log-level of falco to show libs-debug
* Restart containerd on the host
* Logs will show failing containerd-socket tries
* falco will continue to try to reconnect, but it will never work

**Expected behaviour**

Falco has no issue when containerd gets restarted on host

**Additional context**

The issue is caused by mapping the socket as a simple file into the pod. These mappings are done inode-based, so a restarting containerd-process will drop&create a new socket-file on the host, causing a new inode to be used. The inode within the pod will no longer point to the socket-file.

Fix: Instead map the socket-file-containing-folder into the pod, so falco detects the broken socket, retries and is able to use the new inode of the socket.

FYI: A suitable patch to provide this functionality is on the way.

devfaz:
/remove-lifecycle stale

sorry, totally forgot about this issue. Will provide the requested change asap

PR #1058, opened 2026-09-05
fix(charts/falco): mount the container engine socket directory, not the socket

### LINKED ISSUE #3979: Container plugin sockets are mounted as files again, Falco loses containerd after a runtime restart (regression of #632)
**Describe the bug**

Since the container plugin volumes were introduced (092de0da, chart 5.x+), every socket in `collectors.containerEngine.engines.*.sockets` is mounted into the pod as a file hostPath. A file bind mount pins the inode. When containerd restarts on the host it unlinks and recreates the socket, and the pod keeps the old dead inode. From that point Falco cannot reach the runtime until the pod is restarted, and all events for containers created after the restart come out with `container.name=<NA>`, `container.image.repository=<NA>`, `k8s.ns.name=<NA>`, `k8s.pod.name=<NA>`.

This is exactly falcosecurity/charts#632, which was fixed in falcosecurity/charts#633 by mounting the parent directory instead of the file. The fix was lost in the rewrite.

**How to reproduce it**

1. Install chart 9.1.0 (Falco 0.44.1, container plugin 0.7.1) on a node running containerd. In my case RKE2 v1.36.1, socket `/run/k3s/containerd/containerd.sock`.
2. Start a pod, check that Falco events carry `container.name` and `k8s.pod.name`.
3. Restart the runtime on the host (`systemctl restart rke2-server`, or containerd).
4. Start another pod, run something that matches a rule. All container and k8s fields are `<NA>`. Containers created before the restart still get enriched (already cached).

Inode check on my node, two months after the last restart of the Falco pod:

```
# inside the falco container
stat -c '%i %y' /host/run/k3s/containerd/containerd.sock
1973859 2026-06-21 15:32:18

# on the host
stat -c '%i %y' /run/k3s/containerd/containerd.sock
59278913 2026-08-21 06:59:22
```

**Expected behaviour**

Falco keeps working across a runtime restart, like it did after falcosecurity/charts#633. Mount the socket's directory (`dir $socket`) and point the plugin at the file inside it.

**Additional context**

- Runtime restarts are routine on Ubuntu nodes: `unattended-upgrades` + `needrestart` bounce `rke2-server` after any libc/libcurl upgrade. On my cluster that happened 35 times in two months, so enrichment has been broken since June 30 and nothing in the chart signals it (pod Running, `/healthz` fine, metrics fine).
- Even with a directory mount the container plugin does not re-subscribe to CRI events after the stream dies. I opened a separate issue on falcosecurity/plugins for that: falcosecurity/plugins#1499.
- Side effect of the file mounts: the five sockets that don't exist on the node (`/var/run/docker.sock`, `/run/podman/podman.sock`, `/run/crio/crio.sock`, `/run/containerd/containerd.sock`, `/run/host-containerd/containerd.sock`) get created as empty directories on the host by kubelet because the hostPath has no `type`.

Chart 9.1.0, Falco 0.44.1, container plugin 0.7.1, RKE2 v1.36.1 (containerd v2.2.3-k3s1), Ubuntu 24.04.

arpitjain099:
**What type of PR is this?**

/kind bug

/area falco-chart

`falco.containerPluginVolumes` mounts each entry of `collectors.containerEngine.engines.*.sockets` as a file hostPath. A file bind mount pins the inode, so when the runtime restarts and recreates its socket the pod keeps the old dead one. Falco cannot reach the runtime after that, and every container created since the restart reports `container.name`, `container.image.repository`, `k8s.ns.name` and `k8s.pod.name` as `<NA>` until the pod is restarted.

This is the same failure as falcosecurity/charts#632, which falcosecurity/charts#633 fixed by mounting the parent directory. The fix was lost when the container plugin volumes were introduced.

Both helpers now mount `dir $socket` and mount it at `/host<dir>`, so `/host<socket>` still resolves for the plugin config, which passes the socket paths unchanged. Deduplication moves to the directory, so two sockets in one directory produce one volume rather than two mounts on the same path.

The rendered default output is six directories, one per engine socket: `/var/run`, `/run/podman`, `/run/host-containerd`, `/run/containerd`, `/run/crio` and `/run/k3s/containerd`. Volume names and their order are unchanged.

Fixes falcosecurity/falco#3979

Chart version bumped to 9.1.1 for `ct lint --check-version-increment`, with a CHANGELOG entry under Unreleased.

The unit tests in `tests/unit/containerPlugin` asserted the socket file paths, so their expectations move to the directories. Running them against the current template with the updated expectations fails as you would want:

    expected: "/var/run"
    actual  : "/var/run/docker.sock"

With the template change `go test ./unit/containerPlugin/` passes, and `helm template charts/falco` renders the six volumes and matching mounts above.


leogr:
Hey @arpitjain099

Thanks a lot for this fix, and good catch on the regression 🙏

As the bot said above, the Falco chart source now lives in the falcosecurity/falco repository under `chart/falco`, and this repository only receives the synced copy. So I've re-filed your commit there (as-is, your authorship is preserved) 👉 https://github.com/falcosecurity/falco/pull/3978

The only differences: no `Chart.yaml` bump (the version is bumped by the chart release PR only), the changelog entry is under `## Unreleased`, and there is one extra unit test case for two sockets in the same directory.

I'll close this PR once the new one is merged. The fix will ship with the next chart release (Falco 0.45.0).

Thanks again 🙏

leogr:
Hey @arpitjain099

The re-filed PR has just been merged 👉 https://github.com/falcosecurity/falco/pull/3978

Your commit landed as-is, with your authorship, so I'm closing this one. The fix will ship with the next chart release, alongside Falco 0.45.0.

Thanks again 🙏

PR #1059, opened 2026-09-07
feat(falco-talon): mount rules ConfigMap as a directory for hot reload

gitlayzer:
**What this PR does / why we need it:**

kubelet never propagates ConfigMap updates to `subPath`-mounted files, so with the current mount (`mountPath: /etc/falco-talon/rules.yaml` + `subPath: rules.yaml`) **no filesystem event is ever produced** when the rules ConfigMap changes. The `watch_rules` feature can therefore never reload in practice, and users must restart the pods for rule changes to take effect.

This PR mounts the rules ConfigMap as a directory at `/etc/falco-talon/rules.d` and points `-r` at the file inside it. ConfigMap volume updates (symlink swap) then propagate and the watcher reloads rules on the fly, with zero pod restarts.

**Companion to** falcosecurity/falco-talon#799, which fixes the watcher itself to catch the swap events — both are needed for hot reload to work end-to-end.

**Testing performed:**

- `helm lint` passes, `helm template` renders the new args/mount correctly
- on a 3-node kubeadm cluster (chart installed with this change + watcher fix from falco-talon#799): editing the rules ConfigMap → talon logs `changes detected` + `N rules have been successfully loaded` within ~20s, **no pod restarts**; a matching falco event then triggers the freshly loaded rule's action successfully

**Chart version:** bumped to 0.5.0 with CHANGELOG entry.

Signed-off-by: gitlayzer <mr-xintian-du@gmail.com>

### DIFF
--- a/charts/falco-talon/CHANGELOG.md
+++ b/charts/falco-talon/CHANGELOG.md
@@ -6,6 +6,7 @@ numbering uses [semantic versioning](http://semver.org).
 ## 0.5.0 - 2026-09-07
 
 - mount the rules ConfigMap as a directory (`/etc/falco-talon/rules.d`) instead of a `subPath` file, so ConfigMap updates are propagated by kubelet and picked up by the rules watcher (`watch_rules`) — see falcosecurity/falco-talon#799
+- bump `appVersion` to `0.3.1` (the falco-talon release containing the watcher fix) so the rendered image tag matches the hot-reload behavior
 
 ## 0.4.1 - 2026-06-15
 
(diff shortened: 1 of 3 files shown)


Issif:
note : this PR introduces a breaking change and will be merged only after the next release of Talon

gitlayzer:
Closing this PR — the companion talon PR (falcosecurity/falco-talon#799) has been closed, and we've decided to keep the feature for internal use at this time. The directory-mount change stays available at gitlayzer/charts (feat/talon-rules-dir-mount) if anyone wants to pick it up later.

